PR #17, opened 2017-02-10
validator_derive: Properly skip foreign attributes.

theduke:
The old code only skips attributes if they are MetaItem::List.

This was interfering with other struct level attributes for me.

With a simple change, all attributes are skipped if they are not named "validate".


theduke:
Weird, tests pass just fine when I run them locally.

theduke:
Ah alright, tests are failing on newest nightly, doesn't have anything to do with my changes.

Keats:
Thanks!

PR #18, opened 2017-02-16
Implement Error for Errors.

theduke:
For Dispaly, it's not the prettiest output imaginable,
since it just contains error codes.

But it's still better than nothing.


Keats:
Sorry, completely missed that!

Issue #10, opened 2017-01-17
Define a format for errors and/or allow user defined error messages


Keats:
Right now it only returns the error code for the validation that failed, which is enough for me as I usually have control of both frontend and backend and just want to know which bit failed on which field.

However, having some more verbose error or user-defined ones (or both) would be welcome. 
For example when sending error message to a third party doing an API calls or if you don't control the frontend and need to send all the info (args used for the validator) to it.

A few things to consider:

### Serialize validators
Change the error type of `validate` from `Result<(), HashMap<String, Vec<String>>>` to `Result<(), HashMap<String, Vec<Validator>>>` and have `Validator` manually (so we can add their name) implement `Serialize` from Serde.

**Pros**
That's the easy option since this way we will get the name of the validator and all its args easily
**Cons**
Doesn't really work if we want custom error messages

### Custom message
Add an optional `error_msg` field to every single validator that will be sent with the error code. Sending only the error message makes it less flexible wrt if the frontend wants to change it, so we should it still have both error code and error. Error type will be `Result<(), HashMap<String, Vec<(String, Option<String>)>>>` 

### Default error messages
A neat way to do that would be to impl a function on Validator that returns a default error message. I would ignore i18n in default error messages as I don't think it will provide a good end user interface since end users will end up with errors like "Bank account number must match [0-9]{7}" or something like that that I got today with Transferwise. 
Proper error messages should be built by using the serialized validators in the frontend or just overwritten in the backend directly. There are also too many format for i18n and afaik not library for it in Rust yet.

The user could iterate on the errors hashmap or we could even impl a function on the hashmap directly so the user can do something like if they just want a hashmap with error messages:

```
match signup_data.validate() {
   Ok(()) => (),
   Err(e) => return e.get_error_messages()
};
```
which could be turned into a `validate_or_error_messages!` macro easily

If that approach works, it could be extended to transform errors into i18n objects potentially

## Current thoughts
I like the idea of serializing validators and having default error messages if wanted but I'm not sure where the custom error messages fit in that system. Ideally we'd have `String | Validator` as type for the value of hashmap but that's not possible in Rust. Another option is `(Validator, Option<String>)`.
The main issue would be what to do with the `custom` validator since you can have multiple kind of errors and how to work with serialization errors if we decide to let validator handle those (for consistent errors)

theduke:
My proposal would be to add a new trait and common struct for validations.

I also need my validation errors to carry additional custom data, for i18n.

[...]

PR #26, opened 2017-05-24
Feature/error messages partial fix for #10

Leopard2A5:
Partial fix for #10 
Sorry it's a bit lengthy, but it wasn't easy to break this up!

### DIFF
--- a/validator/src/contains.rs
+++ b/validator/src/contains.rs
@@ -40,8 +40,12 @@ impl<'a, S> Contains for &'a HashMap<String, S> {
 /// Validates whether the value contains the needle
 /// The value needs to implement the Contains trait, which is implement on String, str and Hashmap<String>
 /// by default.
-pub fn validate_contains<T: Contains>(val: T, needle: &str) -> bool {
-    val.has_element(needle)
+pub fn validate_contains<T: Contains>(val: T, needle: &str) -> Result<(), String> {
+    if val.has_element(needle) {
+        Ok(())
+    } else {
+        Err(format!("must contain '{}'", needle))
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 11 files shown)


validator/src/contains.rs
```diff
@@ -40,8 +40,12 @@ impl<'a, S> Contains for &'a HashMap<String, S> {
 /// Validates whether the value contains the needle
 /// The value needs to implement the Contains trait, which is implement on String, str and Hashmap<String>
 /// by default.
-pub fn validate_contains<T: Contains>(val: T, needle: &str) -> bool {
-    val.has_element(needle)
+pub fn validate_contains<T: Contains>(val: T, needle: &str) -> Result<(), String> {
+    if val.has_element(needle) {
+        Ok(())
+    } else {
+        Err(format!("must contain '{}'", needle))
```
Keats: I don't want the messages to be in the `validator` crate, they should be in the `validator_derive`.
Validator fns should just return a bool

validator/src/types.rs
```diff
@@ -49,6 +59,39 @@ impl std::error::Error for Errors {
     }
 }
 
+#[derive(Debug, PartialEq)]
+pub struct Error {
```
Keats: `Error` -> `ValidationError`?

validator/src/types.rs
```diff
@@ -49,6 +59,39 @@ impl std::error::Error for Errors {
...
+    validator: String,
```
Keats: `validator` -> `code`

validator/src/types.rs
Leopard2A5: I kept with your approach, the Errors struct isn't called ValidationErrors, so why add the prefix for the singular struct?

Leopard2A5:
Well if you want to implement it yourself just say it and don't merge this PR.

validator/src/types.rs
Keats: It clashes with `std::error::Error`, at least in my mind. But after thinking about it, it's not that bad

Keats:
Sorry about that :(

Keats:
Closing in favour of #27

Issue #22, opened 2017-04-27
serde: support for validation during deserialization


lucab:
This is basically cross-linking the whole discussion at https://github.com/serde-rs/serde/issues/642.

It would be nice to have this crate and serde team-up to provide some way for performing a validation pass as the last step of deserialization. For simple average-case scenarios, I'd like to avoid explicitely calling `validate()` on a struct that has just been deserialized by serde.

The usecase for this is to automatically ensure that a struct is guaranteed to be either valid right after deserialization or to fail it, in order to avoid invalid states between deserialization and validation (or worse, missing validation at some call sites).

Keats:
Ah yes that would be a cleaner approach than what I was thinking with #11 

I'm pretty busy with Tera and Gutenberg currently but I'll resume work on that once Gutenberg gets closer to 0.1

Ideally I'd also like as much feedback as possible on #10 before starting up again to ensure I'm not missing any usecase and have to rewrite the whole thing

ckaran:
Has there been any progress on this?

Keats:
serde doesn't have hooks for that so no

ckaran:
Are there any plans to add in the necessary hooks?

Issue #20, opened 2017-04-02
Use `Cow<'static, str>` instead of `String`


alexander-irbis:
Hi!

There is no point to allocate memory each time for a `String` value, much frequently we will need an `&'static str` value and only in rare cases we will need to allocate `String` for custom values.

for `validate`:
```rust
Result<(), HashMap<String, Vec<String>>>
```
=>
```rust
Result<(), HashMap<Cow<'static, str>, Vec<Cow<'static, str>>>>
```
or even
```rust
Result<(), HashMap<&'static str, Vec<Cow<'static, str>>>>
```

for custom validation functions respectively:
```rust
Option<Cow<'static, str>
```

Keats:
The whole API is probably going to completely change with #10 but I haven't had the time/wanted to work on this project recently so it will take some time.

Keats:
In 0.6.0

PR #27, opened 2017-05-28
Complete refactor

Keats:
This allows #10 in a basic but effective way: the `validator` crate exposes a `ValidationError` and `ValidationErrors`.

`ValidationError` is identical with the one with the same name in Django (code, message and params) and `ValidationErrors` is just a HashMap<FieldName, Vec<ValidationError>>.

In addition to the current validation parameters, each validator now allows to set a `code` and a `message` directly from the attribute. For example:

```rust
#[derive(Debug, Validate)]
struct TestStruct {
    #[validate(regex(path = "RE2", code = "oops"))]
    val: String,
}

#[derive(Debug, Validate)]
struct TestStruct {
    #[validate(email(message = "oops"))]
    val: String,
}
```

The `params` will automatically be filled for each field validation with:

- the `value` of the field
- the parameters to the validator, when it make sense (eg path to a function or regex are useless to write an error message). For example the `range` validator will always have 3 params: `value`, `min` and `max`

Most of the `String` types have been switched to `Cow<'static, str>` as well (#20) 

Things left to think about:

- don't use serde_json: find a cross data format way to store them in the struct
- bikeshedding on the attributes names (https://github.com/Keats/validator/blob/revamp/validator_derive/src/lib.rs#L317-L331)
- renaming Custom validator to Function? more bikeshedding basically
- Use a trait instead of a struct for ValidationError? (see #10)
- How to extract serde serialization errors? (#22)


Keats:
@theduke @alexander-irbis @Leopard2A5 

Going to wait for some feedback now. If you see things to change/improve, please do a PR on that branch.
Documentation hasn't been updated yet in case there are more changes coming like using traits.

Keats:
And @lucab for #22

theduke:
Sorry didn't get around to checking it out. 

I will this evening.

Keats:
I'm updating the docs on master, I'll push in a bit

Keats:
@theduke did you have time to look at it? I've seen that https://github.com/dtolnay/erased-serde got a new version so maybe it can used instead of serde_json

PR #30, opened 2017-08-27
Fix a small typo in changelog dates

tomhoule:
Since 0.6.0 is already on crates.io I suppose it was supposed to be this month and not last month.

Love the library by the way :)


Keats:
Oops, thanks!

PR #32, opened 2017-09-21
Implement Error and Display for ValidationErrors

### LINKED ISSUE #29: Error and Display are not implemented for validator::ValidationErrors
Which makes it quite inconvenient and unintuitive to use the result of `Validate::validate` with the `?` operator. While a trivial implementation of `std::error:Error` would probably suffice, I'm not sure about the implementation of `std::fmt::Display`.

There also seems to be a mistake in README.md:
```rust
let errors = signup_data.validate()?;
```
According to https://github.com/Keats/validator/blob/1c0f14aaec42b772fa22861aba8c8b80fcda5b8a/validator/src/traits.rs#L81
`validate` doesn't return anything.

voidxnull:
Fixes #29.

I couldn't come up with a better implementation of `fmt::Display` so it's just `fmt::Debug` for now.

### DIFF
--- a/validator/src/types.rs
+++ b/validator/src/types.rs
@@ -59,3 +59,14 @@ impl ValidationErrors {
         self.0.is_empty()
     }
 }
+
+impl std::error::Error for ValidationErrors {
+    fn description(&self) -> &str { "Validation failed" }
+    fn cause(&self) -> Option<&std::error::Error> { None }
+}
+
+impl fmt::Display for ValidationErrors {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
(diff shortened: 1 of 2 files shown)


Keats:
Can you add a test that uses `?` on `validate`? Unless my memories are wrong, I don't think it's that easy

voidxnull:
Done. I'm not sure it's needed though. Basically, I just wrote a test for capabilities of the language. `?` works exactly like `try!`. Simple test: https://play.rust-lang.org/?gist=f7f31467ca0034bf9940c33af2db2ab4&version=stable.

Keats:
Hmm I thought I had some issues with `?` when I tried, oh well thanks!

PR #33, opened 2017-09-26
Add a credit card number validator

### LINKED ISSUE #15: Add credit card validator
http://docs.pylonsproject.org/projects/colander/en/latest/extending.html#defining-a-new-validator

Eijebong:
Instead of using our own algorithm, I chose to use rs-card-validator and
put it behind the `credit_cards` features

Fixes #15


validator/src/validation/cards.rs
```diff
@@ -0,0 +1,29 @@
+use validation::Validator;
+use card_validate::Validate;
+
+
+pub fn validate_credit_card(card: &str) -> bool {
+    println!("{:?}", card);
```
Keats: you can remove that

validator/Cargo.toml
```diff
@@ -16,3 +16,7 @@ idna = "0.1"
 serde = "1"
 serde_derive = "1"
 serde_json = "1"
+card-validate = { version = "1.0", optional = true }
+
+[features]
+credit_cards = [ "card-validate" ]
```
Keats: looking at `card-validate` dependencies, I think it's fine to depend on it directly without the features, it only adds one dep

validator/src/validation/cards.rs
```diff
@@ -0,0 +1,29 @@
...
+    match Validate::from(card) {
```
Keats: I think `Validate::from(card).map(|v| v.valid).unwrap_or(false))` would be cleaner

validator/src/validation/cards.rs
Eijebong: Woups, had a test failing and forgot to remove this line, sorry...

Eijebong:
@valeriansaliou You should have released that as a 2.0 (it's a breaking change) :) Can you do that and [yank](http://doc.crates.io/crates-io.html#cargo-yank) that 1.1 ?

valeriansaliou:
@Eijebong just did it thanks.

Keats:
Oops sorry, forgot about that

PR #34, opened 2017-10-10
Add a phone number validator

valeriansaliou:
This validator can be handy when writing APIs, where the front-end does the work of retrieving a country and national phone number, and generating an international phone number (eg. `+33xxxxx`, `+1xxxxx`, etc).

I used on the best phone number validation library I could find ([phonenumber](https://crates.io/crates/phonenumber)), which unfortunately has a lot of dependencies. Thus, I added a `phone` feature to `validator`.

Note that it does not work given a national (non-international) number (eg. 06xxxxx w/o the country indicator), as the library cannot figure out the country by itself to validate the local number (it can be provided to `phonenumber `, but it's a bit of an overhead for `validator` purposes here).

The unit tests are all green for the `validator` part. However, I couldn't figure how to properly test /write the `validator_derive` part, knowing `phone` is an optional feature. Any hint on this? 🙂

I did a first commit for `validator`, and a second commit for `validator_derive` to separate things properly for your review.

Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,17 +6,17 @@ Macros 1.1 custom derive to simplify struct validation inspired by [marshmallow]
 [Django validators](https://docs.djangoproject.com/en/1.10/ref/validators/).
 It relies on the `proc_macro` feature which is stable since Rust 1.15.
 
-By default all args to a `validate` must be strings if you are using stable. 
+By default all args to a `validate` must be strings if you are using stable.
 However, if you are using nightly, you can also activate the `attr_literals` feature to be able to use int, float and boolean as well.
 
 
 A short example:
 
 ```rust
-#[macro_use] 
+#[macro_use]
 extern crate validator_derive;
 extern crate validator;
-#[macro_use] 
+#[macro_use]
 extern crate serde_derive;
 extern crate serde_json;
 
@@ -27,6 +27,8 @@ use validator::{Validate, ValidationError};
 struct SignupData {
     #[validate(email)]
     mail: String,
+    #[validate(phone)]
+    phone: String,
     #[validate(url)]
     site: String,
     #[validate(length(min = "1"), custom = "validate_unique_username")]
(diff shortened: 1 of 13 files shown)


validator/Cargo.toml
```diff
@@ -16,3 +16,7 @@ idna = "0.1"
 serde = "1"
 serde_derive = "1"
 serde_json = "1"
+phonenumber = { version = "0.1.0+8.7.0", optional = true }
```
Keats: that's some weird version number :o

validator_derive/src/validation.rs
```diff
@@ -213,7 +213,7 @@ pub fn extract_argless_validation(validator_name: String, field: String, meta_it
         }
     }
 
-    let validator = if validator_name == "email" { Validator::Email } else { Validator::Url };
+    let validator = match validator_name.as_ref() { "email" => Validator::Email, "phone" => Validator::Phone, _ => Validator::Url };
```
Keats: more readable on multiple lines imo

validator/Cargo.toml
valeriansaliou: Yes, could not figure out why. That library is a Rust port of [https://github.com/googlei18n/libphonenumber](googlei18n/libphonenumber), for which the `8.x.x` part belongs.

validator_derive/src/validation.rs
valeriansaliou: Fixed

README.md
```diff
@@ -151,8 +153,13 @@ Examples:
 #[validate(regex = "ALLOWED_USERNAMES_RE"))]
 ```
 
+### phone
+Tests whether the String is a valid phone number (in international format). To use this validator,
```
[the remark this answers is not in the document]
valeriansaliou: Fixed

valeriansaliou:
Pushed my fix commit, thanks for the review.

Keats:
Sorry, completely forgot about that. Can you rebase?

valeriansaliou:
I've manually merged @Eijebong changes here. Should be no more conflict once you review my latest commits (sorry for the spam!).

Also, I've added the missing `credit_cards` features for @Eijebong card validation PR, the docs states the feature exists but it was never applied in code.

valeriansaliou:
Oh wow okay, though you wanted to minimize the amount of deps as most users won't need those extra features. `credit_cards` feature removed (maybe we could have kept the feature tag and enabled it by default in default features?)

All my commits have been rebased into one. Should be good to go now.

Eijebong:
Github UI is still showing conflicts, so I guess you missed something :p

valeriansaliou:
Sounds good to me, the conflicts are those based on last reviewed commit. My latest commit wasn't yet reviewed.

Keats:
Had to merge it manually but this is in

Keats:
Ok I had to make a few changes for the credit card but I believe everything
on crates.io should be working now

PR #37, opened 2017-12-17
Move extraction of message and code in single function

gtors:
Just for the sake to reduce code duplication.


Keats:
Neat, thanks!

PR #44, opened 2018-03-18
Support for validating Option<Option<T>>. Fixes #41

### LINKED ISSUE #41: Does not work on Option<Option<T>>
Though validation works on `Option<T>`, it would be beneficial if the same logic would work on `Option<Option<T>>`. 
To communicate with the JS world with `null`, `undefined` & `T`, I need to use `Option<Option<T>>`. And to validate it, currently I am using custom validation. 
So, this is a feature request to make it built-in.

csharad:
I read the code and kind of understood :). So, if there are any problems with the PR, hit me up.
Fixes #41

### DIFF
--- a/validator_derive/src/asserts.rs
+++ b/validator_derive/src/asserts.rs
@@ -1,5 +1,5 @@
 
-pub static NUMBER_TYPES: [&'static str; 24] = [
+pub static NUMBER_TYPES: [&'static str; 36] = [
     "usize", "u8", "u16", "u32", "u64",
     "isize", "i8", "i16", "i32", "i64",
     "f32", "f64",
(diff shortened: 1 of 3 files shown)


Keats:
Code looks good but it still looks like using `Option<Option<...>>` is weird. Is it not possible to use an enum like:

```rust
enum JSValue {
  Null,
  Undefined,
  Number,
  String,
  Bool,
  Array<JSValue>,
  Object<JSValue, JSValue>,
}
```

to interop with JS?

csharad:
First, the JSValue would be required to work with `diesel` (impl `ToSql`, `FromSql`) and other such libraries. 
Second,  `JSValue` has a lot of unrequired variants for such a usecase.
Third,  when I would enter/get a custom `T`, I would have to convert it to/from JSValue first, which is just a lot of hoops to jump whereas we would just use `Option<Option<T>>` without any precondition.

Keats:
It's in 0.6.4!

PR #46, opened 2018-04-13
Fix path parsing for regex validators

barzamin:
See also: 8ca1fe94d779dcccf4b3f4c23a08f3ede7e10956, #45.


Keats:
Published in 0.6.5, thanks!

barzamin:
@Keats `0.6.5` doesn't seem to exist: https://crates.io/crates/validator/0.6.5

Keats:
@barzamin It's https://crates.io/crates/validator_derive

The `validator` crate just contain the actual validations, not the proc macro

PR #50, opened 2018-05-16
feature gate card validator

### LINKED ISSUE #49: can we feature gate the card validator?
i'd prefer not to pull in the `card-validate` dependency; i don't need it at all for 99% of the things i use this lib for.

barzamin:
fixes #49


Keats:
I somehow misread and thought the issue was about phonenumber :man_facepalming: 
This is a breaking change so I'll check if there is more work to be done on the library before merging & releasing.

Thanks!

PR #54, opened 2018-07-22
Fixed closing parentheses in README

pJunger:
I just noticed that copy-pasting the examples did not work in some cases as there were some inbalanced parentheses.


Keats:
Thanks!

Issue #55, opened 2018-07-27
Validating `Cow<'a, str>` 


simonsparks:
It is often useful to represent a struct's string fields with the Cow<'a, str> type to handle borrowing efficiently. The pattern is well described in the [From &str to Cow](https://jwilm.io/blog/from-str-to-cow/) article by Jow Wilm.

It would be equally useful if the validators in this library that operate on string types also supported fields declared as Cow<'a, str>.

PR #56, opened 2018-07-27
Added support for validating Cow<'a, str> fields.

simonsparks:
Refactored validate functions to accept parameters that are `Into<Cow<'a, str>>` and added `HasLen` and `Contains` trait implementations for the `Cow<'a, str>` type.
Modified assertion and quoting logic to correctly derive validations for <Cow<'a, str>` fields.
Extended test suite to include examples using this type.

#55

### DIFF
--- a/validator/src/traits.rs
+++ b/validator/src/traits.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::HashMap;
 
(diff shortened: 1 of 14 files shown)


Keats:
Looks good thanks!

simonsparks:
No problem :)

Issue #51, opened 2018-06-07
Bug:  Option<..> with value failing to validate


Dowwie:
In the following example, an incorrect postal_cd is provided.  I expected validate() to raise a ValidationError but instead returned Ok(()).  Is this a bug?

```rust
extern crate validator;
extern crate failure;
#[macro_use] extern crate validator_derive;

use validator::{Validate, ValidationError};
use failure::Error;

#[derive(Validate)]
pub struct Address {
    pub street: String,
    #[validate(length(min = "1", max = "5"))]
    pub postal_cd: String
}

#[derive(Validate)]
pub struct PartyInfo {
    #[validate(length(min = "1", max = "5"))]
    pub first_name: Option<String>,
    #[validate(length(min = "1", max = "10"))]
    pub last_name: Option<String>,
    pub address: Option<Address>
}

fn main() -> Result<(), Error>{
    let addr = Address {street: "123 main st".to_string(),
                        postal_cd: "0703000000000000".to_string()};
        let party = PartyInfo {first_name: Some("Darin".to_string()),
                           last_name: Some("Gordon".to_string()),
                           address: Some(addr)};
        let check = party.validate()?;
    println!("check is: {:?}", check);

    Ok(())
}
```

Keats:
See #31 

I'm not even entirely sure how I could detect that a field is implementing `validate`, I would need to find the struct definition somehow and check the attributes. The second part is easy but the first...
If anyone has an idea, that's definitely something I want

Dowwie:
@Keats I hope you don't mind that [I've asked for help](https://www.reddit.com/r/rust/comments/8pj2qg/a_rust_challenge_for_the_validator_crate/) but really would like to see this limitation overcome for the library. :)

Dowwie:
@Keats someone from Reddit (boscop) mentioned that serde may present a "nesteds" example from which to reference

Keats:
I probably won't have time to work on it in the near future but I will happily take a PR for it. 
I'm not entirely sure what should the output in case of errors in nested structs though

Dowwie:
@simonsparks Thanks for the update!  This is good news!

Issue #31, opened 2017-09-17
Validation of nested structs?


e-matteson:
Let's say I'm deserializing to a parent struct whose fields are other structs, which implement Validate. I hoped that if I called `validate()` on the parent, it would automatically call `validate()` on its nested structs. But this code prints out "validated!", even though Bar is invalid: 

```
#[macro_use]
extern crate serde_derive;
extern crate toml;
#[macro_use]
extern crate validator_derive;
extern crate validator;

use validator::{Validate};

#[derive(Deserialize, Debug, Validate)]
struct Foo {
    bar: Bar,
}

#[derive(Deserialize, Debug, Validate)]
struct Bar {
    #[validate(range(min = "0", max = "9"))]
    y: i32
}

fn main() {
    let foo: Foo = toml::from_str(r#"
    [bar]
    y = -1
    "#).unwrap();

    match foo.validate() {
        Ok(_) => println!("validated!"),
        Err(e) => println!("{:?}", e),
    };
}
```  

I can't figure out a  good way to explicitly call `validate()` on the nested struct, either. I tried adding a `#[validate(custom("Bar::validate"))]` attribute to the `bar` field, but that resulted in `error: proc-macro derive panicked`. 

Is there any better way to validate nested structs than to implement my own method for Foo that manually calls `validate()` on every field?

Keats:
I have to say I didn't think of nested structs at all :(

I guess we can modify the proc macro to check if there is a `Validate` tag but that doesn't really work in terms of types for the errors: right now it's a hashmap string -> Vec<Error> but for nested structs we would want hashmap -> (hashmap -> Vec<Error>) to represent errors on various fields. Not sure what's the best plan for that, maybe have an enum to represent the kind of error?
Reference: http://marshmallow.readthedocs.io/en/latest/nesting.html

greglearns:
Could you continue to use hashmap -> Vec if you treated each string key as a "path.to.each.nested.object", so that you could have "top.number" -> "is invalid" and "top.address.street" -> "is missing a street number" or something like that?

Keats:
It would work but it would be a pain to iterate through the errors in the frontend as you would need to check what the field name starts with. Not a proper solution imo

Keats:
If anyone wants to try it before I release a new version, they can use the crates from the `master` branch. Release will probably happen early next week, I want to play with it a bit as well.

simonsparks:
:tada:

PR #60, opened 2018-08-30
Nested Validation

simonsparks:
I've made an implementation of nested struct validation. It uses plain `[#validate]` attributes added to a parent struct's fields in order to detect which nested structs should be validated.
With specialization, it could work without the field attribute as we could speculatively validate all fields, having a default no-op trait implementation for cases where a struct has no real validations to execute. But currently this would have required using nightly so I opted for the explicit attribute approach instead. Also, this gives the user more control over which sub-structs to opt-in for validation.

I aimed to avoid changing the API of the existing library; the main observable difference I think is that the ValidationErrors HashMap is now keyed with a `String` instead of a `&' static str`. I don't know if this will make much difference in real world usage; it didn't cause existing tests to fail at least.
The ValidationErrors keys use a `path.to.nested.field` format to indicate the position of invalid nested fields, and each ValidationError instance includes that path as a `["path", "to", "nested", "field"]` vector for convenience.

This implemetation also works with Optional nested structs and Vectors of nested structs (but currently not combinations of the two).

Keats/validator#31
Keats/validator#51

### DIFF
--- a/validator/src/types.rs
+++ b/validator/src/types.rs
@@ -8,7 +8,7 @@ use serde::ser::Serialize;
 
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
 pub struct ValidationError {
-    pub path: Vec<Cow<'static, str>>,
+    pub path: Vec<String>,
     pub code: Cow<'static, str>,
     pub message: Option<Cow<'static, str>>,
     pub params: HashMap<Cow<'static, str>, Value>,
@@ -24,8 +24,8 @@ impl ValidationError {
         }
     }
 
-    pub fn set_path(mut self, path: Vec<&'static str>) -> Self {
-        self.path.extend(path.into_iter().map(Cow::from));
+    pub fn set_path(mut self, path: Vec<String>) -> Self {
+        self.path.extend(path.into_iter().clone());
         self
     }
 
@@ -46,7 +46,7 @@ impl std::error::Error for ValidationError {
 }
 
 #[derive(Debug, Serialize, Clone, PartialEq)]
-pub struct ValidationErrors(HashMap<String, Vec<ValidationError>>, Vec<&'static str>);
+pub struct ValidationErrors(HashMap<String, Vec<ValidationError>>, Vec<String>);
 
 
 impl ValidationErrors {
@@ -74,7 +74,7 @@ impl ValidationErrors {
     }
 
     pub fn add(&mut self, field: &'static str, error: ValidationError) {
-        let path = FieldPath::concat(Some(field), Some(&self.1));
+        let path = FieldPath::concat(Some(String::from(field)), Some(&self.1));
         self.0.entry(path.join(".")).or_insert_with(|| vec![]).push(error.set_path(path));
     }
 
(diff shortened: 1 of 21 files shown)


Keats:
Amazing!
I was planning to do it myself next week so you saved me some work. I still want to think on how I would represented nested errors without being too influenced so I will review this PR next week after spending some time pondering.

simonsparks:
Sure, no problem.

dtolnay:
I can see why the field attributes are necessary. I don't know of a better way unless you use specialization or require that every field has a Validate impl.

simonsparks:
Yes, specialization works nicely for this feature but of course that would require nightly for now.
I take the requirement for the more structured validation errors though and I'll refactor the PR to reflect that.
Another scenario to consider for this enum is the case where a field has both its own validations and the nested validations of its children, e.g. a vector with a length validation containing validating structs.

Keats:
> Another scenario to consider for this enum is the case where a field has both its own validations and the nested validations of its children, e.g. a vector with a length validation containing validating structs.

It should be the Vec kind at that point, I think.

From marshmallow, if I change a line to`authors = fields.Nested(AuthorSchema, many=True, validate=validate.Length(min=2))` and try it with with

```python
bowie = dict(name="Bowie")
album = dict(artist=bowie, title="Hey", authors=list([dict(name="hey")]))

try:
    schema.load(album)
except ValidationError as err:
  pprint(err, indent=2)
```
I get : `ValidationError({'authors': ['Shorter than minimum length 2.']})` which is what I expected

Keats:
That sounds like a good idea.
More importantly, TIL about `#[deprecated`. On my way to use it everywhere!

Keats:
Thanks, I'll have a look this week

validator_derive/tests/complex.rs
```diff
@@ -1,3 +1,5 @@
+#![feature(box_patterns)]
```
Keats: can we do without that?

validator_derive/tests/contains.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(deprecated)]
```
Keats: why do we need that in many test files?

validator/src/types.rs
```diff
@@ -39,25 +39,102 @@ impl std::error::Error for ValidationError {
...
+        match child {
+            Ok(()) => parent,
+            Err(errors) => parent.and_then(|_| Err(ValidationErrors::new())).map_err(|mut parent_errors| {
+                parent_errors.add_nested(field, ValidationErrorsKind::Struct(Box::new(errors)));
+                parent_errors
+            })
+        }
+    }
+
+    pub fn merge_all(parent: Result<(), ValidationErrors>, field: &'static str, children: Vec<Result<(), ValidationErrors>>) -> Result<(), ValidationErrors> {
```
Keats: can you add docstrings for `has_error`, `merge` and `merge_all`?

Keats:
Ah and can uyou pdate the README with the nested doc as well?

validator_derive/tests/contains.rs
simonsparks: I mentioned the other day that _"all tests currently on the master branch would pass without modification"_. This is true but, because the `inner` function is now marked as deprecated, the test run logs warnings every time it is used.
I added the `allow` tags as I didn't know if Travis would fail the build with warnings but am happy to remove them if this is not the case.
The alternative is to update the tests to use the new `field_errors` method instead.
Any preference?

simonsparks:
Ok, I'll aim to add some nested documentation the README later today.

validator_derive/tests/contains.rs
Keats: I'm tempted to just release a new major version so we can probably get rid of `inner` completely

simonsparks: Your call :)

simonsparks: Perhaps I'll update the test to use the new method and strip out the allow tags anyway.
Then you can decide what version to release either way.

simonsparks:
I updated the README earlier today; I think that's everything?

Keats:
That's perfect! :+1:

PR #62, opened 2018-09-30
Almost done WIP for adding ability to specify multiple schema-level validation functions.

MartinHowarth:
As the title says, this PR adds the ability to specify multiple schema level validator functions. This allows for better separation of validation functionality, compared to a series of statements in a single function.

Example (or see the test below)
```
#[derive(Debug, Validate)]
#[validate(schema(function = "schema_fn"))]
#[validate(schema(function = "schema_fn2"))]
struct TestStruct {
    val: String,
}
```

There's one outstanding issue (see the TODO) which I'll look into in the next week or two, but if you have a recommendation on how to solve it please let me know. I'm new to rust and used this as a learning project - was on the look out for a rust equivalent to the pytohn library [schematics](https://github.com/schematics/schematics).

Let me know what you think :)

### DIFF
--- a/validator_derive/src/lib.rs
+++ b/validator_derive/src/lib.rs
@@ -67,7 +67,7 @@ fn impl_validate(ast: &syn::DeriveInput) -> proc_macro2::TokenStream {
         }
     }
 
-    let schema_validation = quote_schema_validation(find_struct_validation(&ast.attrs));
+    let schema_validations = quote_schema_validation(&find_struct_validation(&ast.attrs));
 
     let ident = &ast.ident;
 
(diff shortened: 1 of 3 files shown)


Keats:
> I'm new to rust and used this as a learning project - was on the look out for a rust equivalent to the pytohn library schematics.

Interesting, I didn't know about schematics but validator is inspired by [marshmallow](https://marshmallow.readthedocs.io/en/3.0/).

What is the usecase of having several functions instead of a single one though?

MartinHowarth:
The use case is simply for keeping concepts separated into different functions. If I have one schema where I need to validate two different things at the schema-level then I'd want to have one function for each of those, rather than having a single function handling all of the validation. Overall makes the resulting code more readable, testable and re-usable.

Keats:
@MartinHowarth are you still planning to finish that PR?

Keats:
@MartinHowarth ping

Keats:
Out of date.

Issue #69, opened 2018-11-09
Allow range validator to only have 1 argument


mehcode:
I'm trying to have just a "greater than" validation (must be `>` than `0`).

```rust
#[validate(range(min = "1"))]
```

Fails with

```
Validator `range` requires 2 arguments: `min` and `max`
```

Keats:
It does look like my memory was wrong, I thought it was possible :(
I am not against the idea but how to define what are the min/max values? Should it use the Rust defaults like `u8::MIN`, `u8::MAX` for each type?

PeteJodo:
could the `min` and `max` fields in `Validator::Range` just be `Option<f64>` and then check only what needs to be checked in the `validate_range` function? You could then have `validator_derive` fail if both are not set. Also irrelevant, but why `f64` and are there any implications with `f64`?

PR #71, opened 2019-04-28
Issue #69: Change `range` to require only one arg

cazgp:
Currently range is hard-coded to take both max and min. This is
unhelpful in cases where we may only want to check one of the bounds.

Update so that it behaves more like the `length` validator and can take
either max or min, as well as both. Also ensure that at least one of
them is supplied.


Keats:
Change looks good! I'll have another look later this week and merge it

PR #79, opened 2019-09-06
Add utf-8 non-control characters validator

mocsy:
Rust Strings are utf-8 and it has many control characters, which are sometimes good to test for.
https://bugzilla.mozilla.org/show_bug.cgi?id=968576

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,11 @@ Examples:
 #[validate]
 ```
 
+### non_control_character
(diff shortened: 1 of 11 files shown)


Keats:
Sorry for the late comment!
I think I would prefer it being called `non_control_character` just to be clearer since `non_control` is not obvious at first sight. 
The PR itself is perfect 👍

Keats:
Thanks!

PR #91, opened 2019-11-13
Update struct level validation in README

Beskhue:
Struct-level validation has been changed in b31fd25cc5cec96ccee737ba1313c6e9f702f32a, but it appears the README was not completely updated: the return value for struct-level validation refers to the old functionality.


Keats:
Thanks!

PR #94, opened 2020-02-15
Add `must_use` attribute to functions returning `bool`

AndiDog:
This leads to warnings if the return value isn't used. I found this particularly helpful in languages like C++, where returning `bool` is much more prone to human misuse than throwing an exception.


Keats:
Thanks!

PR #97, opened 2020-04-10
Specify static lifetime for field_errors

barskern:
Noticed that this could be a `'static` lifetime when using this library in my webserver.


barskern:
Seems like the CI failures are unrelated to this PR.

Keats:
Thanks!

Issue #99, opened 2020-05-13
Add examples with message and code attibutes to the README.md


xFrednet:
I wanted to add a message to a regex validation using a macro. Sadly, I had to look into the tests to find an example how to do this. It would be nice to add examples like:

```Rust
#[validate(regex(path="USERNAMES_RE", message="invalid_username"))]
```

The same would be useful for custom validation. :)

I would volunteer to add a couple of examples, however I wanted to get your opinion first :)

Keats:
Go for it!

xFrednet:
Awesome! 

Do you know if there is any way to add a message or code attribute to a nested validate? @Keats

PR #100, opened 2020-05-13
Added some examples in the README.md #99

xFrednet:
I've added a few more examples than necessary to the `Message and code` section. Let me know if you have any suggestions or changes you would like to see :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,7 @@ Examples:
 
 ```rust
 #[validate(must_match = "password2")]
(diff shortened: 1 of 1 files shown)


Keats:
Is there a way to add all of those to tests to ensure the examples do work as expected?

xFrednet:
The examples are copied from the related test files. I've only changed the string values. I could also add them as additional tests if you'd like :)

Keats:
Ah that's perfect if they come from tests then :)

Issue #101, opened 2020-05-19
Feature suggestion: impl<T: Validate> Validate for &T


LucasPickering:
It would be nice to have `Validate` implemented for any `&T` if it's implemented for `&T`. Since only a reference is needed for validation anyway, it would make it a little more convient to validate non-Copy types. I can't implement this in my own crate because of the orphan rule. I don't _think_ this would cause any breaking changes, but I'm not quite sure. This would obviously be very easy to write so if it's desirable, I'd be happy to put up a PR with it.

Keats:
Please do a PR!

PR #102, opened 2020-05-20
Add a blanket Validate implementation for references

LucasPickering:
As promised, a blanket impl for `Validate` on references to types that are already `Validate`. I confirmed this is working by testing it in my own crate. Closes issue #101


Keats:
Can you add tests for it? Does it move the value?

LucasPickering:
@Keats just added a test. I wasn't sure where to put it, so let me know if you want it moved. And no, the value doesn't move, it just copies the reference.

Issue #21, opened 2017-04-12
Validating `Option::is_some`


kardeiz:
I haven't used your library yet, but it looks nice&mdash;

However, one of the primary validations I would want to perform on my structs would be `Option::is_some`, basically to check whether a builder-style struct is in a valid state.

It looks like the `custom` validator won't work here, since that is only run on the inner value in the `Option::Some` case.

I understand why you've designed it so validators are only run on options if `Option::is_some`, but it would be nice to have either an `is_some` validator or a custom validator that runs on the raw value.

Keats:
Wouldn't it be solved by having your struct not have an option? the serialization would fail if the field is not there but you wouldn't get a nice error message.
A `required` validation could be added I guess?

kardeiz:
@Keats thanks for the reply. I'm not sure what serialization has to do with this. I'm interested in using this library to validate a builder struct before transforming it to another struct, where the final struct has some conditions on the values (i.e. as supported by your crate). The builder struct may have some fields optional (to provide a nicer builder interface) even when they are not optional in the final struct.

lucab:
As far as I understood, @kardeiz you would do it via [struct-level validation](https://github.com/Keats/validator#struct-level-validation) on the parent struct, checking for `is_some()` on a specific field of it.

timvisee:
I think this can be closed, as #103 is implemented.

PR #103, opened 2020-05-20
Required validation

ShadoySV:
#21 Implementation of required validation. Must it automatically check nested or there must be another validation key like `required_and_check_nested`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,9 @@ Tests whether the String has any utf-8 control caracters, fails validation if it
 To use this validator, you must enable the `unic` feature for the `validator_derive` crate.
 This validator doesn't take any arguments: `#[validate(non_control_character)]`;
 
+### required
+Tests whether the `Option<T>` field is `Some`;
+
 ## Struct level validation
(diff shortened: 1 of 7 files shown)


Keats:
What do you mean by check nested? Do you have an example of such a struct?

ShadoySV:
I mean validation nested `T` in `Option<T>` as `#[validate]` do. There are no such validation in proposed commit.

In my use case, validation nested is not required yet, but it may be useful in cases using `[#validate]`, additionally checking if nested is `Some`.

My use case:
Some object is filled by web interface and one of its fields is an reference to another object (`Option<AnotherObjectRef>`, by default is `None`). That field must be selected and be `Some` for writing to database. It is simple to set attribute `#[required]` instead of defining a function for struct level validation.
```
#[derive(Debug, Serialize)]
struct AnotherObjectRef {
    id: i32,
    name: String,
}

#[derive(Debug, Validate)]
struct SomeObject {
    #[validate(required)]
    val: Option<AnotherObjectRef>,
}
```

Keats:
Ah I see, is it hard to add that functionality?

PR #109, opened 2020-08-25
Reexport validator_derive macros from validator crate

### LINKED ISSUE #58: export validator_derive from validator
Can we do something like the failure crate? This would be a bit more ergonomic to use the lib.

ELD:
This solution is admittedly a little silly, but it's a working attempt at exporting the derive macro from the `validator` crate. I'm happy to modify this in any way that meshes best with how the library should be structured.

I also added a root `Cargo.toml` to turn the repo into a workspace sine the `validator` crate is the main and only entry point to behavior.

The changes here are a little sweeping with regards to the ui tests, so I understand if there's reticence there. I switched from `compiletest-rs` to `trybuild` for its ease-of-use and its ability to use Rust 2018 macro imports transparently (no need for `extern crate`, just `use validator::Validate` works). I also had to drop the pinned version since it looks like an upstream dependency no longer builds on Rust 1.36 (`object` it looks like?).

This should resolve #58.


.github/workflows/ci.yml
```diff
@@ -7,39 +7,36 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        build: [pinned, stable, nightly]
```
Keats, line 10: can you keep the pinned and figure out the earliest version that works? We use that to know if we require a new Rust version.

Keats (changes requested):
There is no way to avoid the 3rd crate is it? :(

.github/workflows/ci.yml
ELD: Sure! Happy to do that 🙂

ELD:
@Keats There may be a way to do it without a third crate. I need to inspect more of what the macro does and I'm not quite to the point of saying that there must be a third crate. I have an idea, but I don't know if it'll work. I'll hack on it a bit in this PR and see if I can work it out.

ELD:
So to elaborate a little more, I see two options in front of us for allowing the derive macro to be exported from the `validator` crate. They are:

1. This solution. It breaks out the `Validator` type to a third crate to break the cyclic dependency. Each crate depends on it, so including `validator_derive` doesn't have the issue it did. This is alright because you'd stop publishing `validator_derive` on Crates.io and you also _wouldn't_ publish `validator_types` since those are private crates for the `validator` crate. It does make development potentially cumbersome as every additional shared dependency between the two crates has to go in that third (`validator_types`) crate.

1. We update the `validator_derive` crate to not use the `Validator` type. I'm not sure the full changes required, yet, but in a cursory reading of the code, we would push a token stream for the `Validator` variant we want on to the internal data structures. This allows us to _not_ have a third crate, but it does decouple the type safety. So if the `Validator` type adds new variants or renames a variant (renaming would likely be a breaking change, adding could be too?), there's a risk of the derive macro falling out of sync with it.

Basically, the crux of the problem is the usage of the `Validator` type in the `validator_derive` crate. I'll see if I can come up with any other options. Rather than force option #1, I figure I would enumerate what I've thought of so we can pick the best solution for the library. Additionally, it might spur on a thought for you to come up with other options as well :)

Keats:
> This is alright because you'd stop publishing validator_derive on Crates.io and you also wouldn't publish validator_types since those are private crates for the validator crate.

They will need to be published to crates.io afaik? Unless you meant they don't need to be published often

ELD:
Oh, yes, you're right. The `validator_types` crate would still need to be published.

Keats:
I think the current PR is fine, I just wished it wasn't such an issue publishing macro crates :(

ELD:
I completely agree. I still think option two is possible, it's just a larger set of changes to the derive macro. I'm happy to sketch it out, if you'd like. It does create the out-of-sync risks I mentioned above, but if we don't anticipate significantly changing the `Validator` enum and its variants, it might work.

ELD:
@Keats Is there anything else you'd like me to do on this PR before approval? Thanks!

Keats:
I think it's all good, i'll have another look this weekend if I find some to review/merge/release

ELD:
Sounds good. Just wanted to check in :)

Keats:
Tested it locally and it's fine! Can you update the README wiht the new installation instructions?

ELD:
I updated the README example to show the `derive` feature. I can add more notes, if preferred.

I also haven't updated the changelog section as I don't know whether this will be a point release for 0.10 or if this will be a new 0.11 release. Let me know and I'd be happy to update it.

PR #110, opened 2020-08-27
Invalide Emails

weichweich:
The email validation accepted a few addresses that are invalid (eg. `.John.Doe@example.com`, `John..Doe@example.com`, `John.Doe@exam_ple.com`).

This PR changes the regex so that those addresses are not accepted.

I also think that `a@atm.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa` should be invalid since there are 64 `a`s?

### DIFF
--- a/validator/src/validation/email.rs
+++ b/validator/src/validation/email.rs
@@ -9,7 +9,7 @@ lazy_static! {
     // Regex from the specs
     // https://html.spec.whatwg.org/multipage/forms.html#valid-e-mail-address
     // It will mark esoteric email addresses like quoted string as invalid
-    static ref EMAIL_USER_RE: Regex = Regex::new(r"^(?i)[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*\z").unwrap();
+    static ref EMAIL_USER_RE: Regex = Regex::new(r"^(?i)[a-z0-9.!#$%&'*+/=?^_`{|}~-]+\z").unwrap();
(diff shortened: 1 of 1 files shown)


Keats:
I'm using the HTML spec regex for emails. If it's good enough for browsers and no one changed it, it's probably fine

weichweich:
> I'm using the HTML spec regex for emails. If it's good enough for browsers and no one changed it, it's probably fine

Ok fair enough. But even then there is a bug. The HTML spec does not accept `John.Doe@exam_ple.com` or `a@atm.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa`. If you have a look at [this line](https://github.com/Keats/validator/pull/110/files#diff-d1b395ac143037e76f9b2a07acc3dc7aR14). The dot should be escaped.

Keats:
I agree but I would prefer finding some "official" regex for it rather than tweaking our own. Do you know a better one?

weichweich:
Ok let's use the [spec that's mentioned in the source code](https://html.spec.whatwg.org/multipage/forms.html#valid-e-mail-address).

But the regex we use differs slightly from the regex shown in the spec. They escape the dot and we don't. I reverted most of my changes, but I escaped the dot in the domain regex. I guess that's fine?

Keats:
Thanks, I'll check the pinned version failure, probably just a matter of rebasing or upgrading the Rust version

PR #114, opened 2020-09-08
Use absolute path for trait impl

### LINKED ISSUE #112: Validate should use absolute type names and not rely on an external import
The following code does not compile unless I also add `use validator::Validate` beforehand.

```rust
#[derive(
    Clone,
    Debug,
    Eq,
    PartialEq,
    :: serde :: Serialize,
    :: serde :: Deserialize,
    :: validator_derive :: Validate,
)]
pub struct Message {
    #[validate(length(min = 1i64, max = 2048i64))]
    pub text: String,
}
```

This should not be necessary if `::validator_derive::Validate` did use absolute type names. I'm using `validator_derive` as part of generated code and therefore don't want `Validate` to become a reserved name.

Is there a way to change the generated code to use only absolute type names and not depend on `use validator::Validate`?

ELD:
Turns out this one was pretty trivial. Just a one-line change in the code generated by the procedural macro! 🎉 

Resolves #112


Keats:
Nice! Thanks

PR #117, opened 2020-09-13
fixed README.md

nujz:


Keats:
Yep that would do it, thanks!

PR #118, opened 2020-09-15
README.md: Use Rust 2018 edition, Validate derive reexport

strohel:
- there is no need for `#[macro_use]`, `extern_crate`
- make use of the fact that `Validate` *attribute macro* is now reexported at the same path as the trait, `validator::Validate`
- `serde_json` was not used in examples, no need to mention it

Note: I haven't actually tried to compile it, but hopefully that's "obvious enough"? :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,13 +17,7 @@ validator = { version = "0.11", features = ["derive"] }
 A short example:
 
 ```rust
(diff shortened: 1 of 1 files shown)


Keats:
Thanks!

PR #121, opened 2020-09-21
asserts: Add support for references to length types

Tuetuopay:
This allows validating a `&'a Vec<...>` rather than only `Vec<...>`. Similarly, this will ignore the lifetime parameter, thus allowing `&'a str` in addition to `&str`.

Example usecase: validating Diesel's `Insertable` structs.

When using diesel, it is common to have the `Insertable` structs to only be references for speed purposes, and it is the canonical way they expose in their getting started docs. This allows validating such a structure:

```rust
#[derive(Insertable, Validate)]
pub struct NewUser<'a> {
    /// User's name
    #[validate(length(min = 3, max = 128)]
    pub name: &'a str,

    /// Mainly spoken languages. Limited to 3 for simplicity for other users
    #[validate(length(min = 1, max = 3))]
    pub languages: &'a Vec<String>,
}
```

### DIFF
--- a/validator_derive/src/asserts.rs
+++ b/validator_derive/src/asserts.rs
@@ -80,6 +80,15 @@ pub fn assert_type_matches(
 }
 
 pub fn assert_has_len(field_name: String, type_name: &str, field_type: &syn::Type) {
+    if let syn::Type::Reference(ref tref) = field_type {
+        let elem = &tref.elem;
+        let type_name = format!("{}", quote::quote!{ #elem }).replace(' ', "");
+
+        if type_name == "str" { return }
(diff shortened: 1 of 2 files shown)


Keats:
Can you add tests for that?

Tuetuopay:
> Can you add tests for that?

I just added one for `&'a Vec<String>`, but if you think others would be useful, don't hesitate!

Tuetuopay:
Hm weird, the test pass locally... Fixing.

Tuetuopay:
@Keats tests have been fixed.

Keats (approved):

Keats:
Thanks!

Issue #7, opened 2016-12-29
Make `must_match` work with Option?


Keats:
Not sure if it even make sense to be honest, I had
```rust
pub fn validate_must_match<T>(a: T, b: T) -> bool
    where
        I: Eq,
        T: Into<Option<I>>,
        T2: Into<Option<I>>,
{
    let a = a.into();
    let b  = b.into();
    if a.is_none() && b.is_none() {
        return true;
    }
    if !(a.is_some() && b.is_some()) {
        return false;
    }
    a == b
}
```
before removing it

Keats:
Probably not needed

Keats:
Resurrected as I'm making validators work on `Option<>` but i don't see a use case for optional `must_match` so I won't implement it without a solid usecase

cazgp:
Hello! I have a potential use-case for this situation.

A user is able to edit their profile. They can optionally set their password. If they set password1, then they must set password2. However, if neither field is provided then they must both be `None`.

The logic is therefore exactly as you've proscribed above.

Keats:
Seems like a legit usecase!

Keats:
@cazgp are you willing to add support for that? It should be fairly easy

cazgp:
Hey! I'm not currently working on this due to some time constraints, but am more than happy to pick it up when getting back round to it... just not exactly sure when that will be :(

Keats:
Don't worry, take your time

xFrednet:
What is the current status on this issue? I would try to implement this if it's still open :)

Keats:
Go for it

xFrednet:
@Keats I've been doing some testing and I believe that this functionality is actually covered by the default `Eq` implementation of `Option`.

I tested the following statements and they where all handled without a problem:
```rust
let a: Option<u64> = None;
assert!(validate_must_match(a, None));
assert!(validate_must_match(Some(6), Some(6)));

assert_eq!(false, validate_must_match(Some(2), Some(3)));
assert_eq!(false, validate_must_match(None, Some(3)));
assert_eq!(false, validate_must_match(Some(3), None));
```

I would simply create a pull request with these tests unless I've misunderstood something?

PR #131, opened 2020-11-21
Added tests for must_match validation with options #7

xFrednet:
I've added some tests to ensure that the `must_match` validation also works with `Option<T>`


xFrednet:
I believe that this should also close the Issue #7

Keats:
Thanks!

PR #132, opened 2020-11-22
The `range` validator can now take a value reference as a parameter #66

### LINKED ISSUE #66: Allow range validator to refer to a variable
The range validator requires the raw numerical value in quotes. It should let it be a variable name in scope as well (or instead).

xFrednet:
This implementation required some changes when it comes to the `Validator::Range` value. I therefor suggest that we add a depreciation notice on that value and add a new value `Validator::RangeRef` for compatibility reasons. This will only effect users that used the validation directly, indirect usage with `#[derive(Validation)]` will be unaffected.  The `validate_range` function was also extended to take generic parameters. The old function still exists and forwards the call to `validate_range_generic`.

These changes are enough to raise the version number, especially with the suggestion that we depreciate something. I therefor noted `0.12.0` as the depreciation version. (I can remove the depreciation notice, that is just a suggestion)

---

This implementation supports constant and struct values like this:

```rust
// Global constant values
#[validate(range(min = "MAX_CONSTANT"))]
#[validate(range(min = "crate::MAX_CONSTANT"))]

// Struct value
#[validate(range(min = "self.max"))] 
```

This separation in struct and global values is readable but brings some inconstancy when it comes to other validations. The `must_match` validator only takes struct values. Changing it to also use the `self.` prefix would be very simple but would break some current usages.

I can easily take this out again but I think that it would be cool if this would be consistent in the entire crate.

---

The validation message resolves the references automatically the error message contains the number it self and not the referenced value: 
```
ValidationErrors({"val": Field([ValidationError { code: "range", message: None, params: {"value": Number(6), "max": Number(5), "min": Number(4)} }])})
```

---

This should also close #66

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,15 +167,18 @@ At least one argument is required with a maximum of 2 (having `min` and `max` at
 Examples:
 
 ```rust
+const MIN_CONST: u64 = 1;
+const MAX_CONST: u64 = 10;
+
 #[validate(length(min = 1, max = 10))]
 #[validate(length(min = 1))]
 #[validate(length(max = 10))]
 #[validate(length(equal = 10))]
+#[validate(length(min = "MIN_CONST", max = "MAX_CONST"))]
 ```
 
 ### range
 Tests whether a number is in the given range. `range` takes 1 or 2 arguments `min` and `max` that can be a number or a value path.
-The range validation can also reference struct values by using `self.` as a value prefix.
 
 Examples:
 
@@ -190,7 +193,6 @@ const MIN_CONSTANT: i32 = 0;
 #[validate(range(max = 10.8))]
 #[validate(range(min = "MAX_CONSTANT"))]
 #[validate(range(min = "crate::MAX_CONSTANT"))]
-#[validate(range(min = "self.max"))] // <- References the value max from the struct
 ```
 
 ### must_match
--- a/README.md
+++ b/README.md
@@ -56,9 +56,10 @@ match signup_data.validate() {
 };
 ```
 
-The `validate()` method returns a `Result<(), ValidationErrors>`. In the case of an invalid result, the
-`ValidationErrors` instance includes a map of errors keyed against the struct's field names. Errors may be represented
-in three ways, as described by the `ValidationErrorsKind` enum:
+A validation on an `Option<_>` field will be executed on the contained type if the option is `Some`. The `validate()`
+ method returns a `Result<(), ValidationErrors>`. In the case of an invalid result, the `ValidationErrors` instance includes
+a map of errors keyed against the struct's field names. Errors may be represented in three ways, as described by the 
+`ValidationErrorsKind` enum:
 
 ```rust
 #[derive(Debug, Serialize, Clone, PartialEq)]
@@ -86,10 +87,6 @@ The value of the field will automatically be added to the params with a key of `
 Note that `validator` works in conjunction with serde: in the example we can see that the `first_name`
 field is renamed from/to `firstName`. Any error on that field will be in the `firstName` key of the hashmap,
 not `first_name`.
-
-If you are adding a validation on a `Option<..>` field, it will only be ran if there is a value. The exception
-being `must_match` that doesn't currently work with `Option` due to me not finding a use case for it. If you have one,
-please comment on https://github.com/Keats/validator/issues/7.
 
 The other two `ValidationErrorsKind` types represent errors discovered in nested (vectors of) structs, as described in
 this example:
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -183,16 +180,23 @@ Examples:
...
+
 #[validate(range(min = 1))]
-#[validate(range(min = 1, max = 10.8))]
+#[validate(range(min = "MIN_CONSTANT"))]
+#[validate(range(min = 1, max = 10))]
 #[validate(range(min = 1.1, max = 10.8))]
 #[validate(range(max = 10.8))]
+#[validate(range(min = "MAX_CONSTANT"))]
+#[validate(range(min = "crate::MAX_CONSTANT"))]
+#[validate(range(min = "self.max"))] // <- References the value max from the struct
```
Keats, line 199: Is that something someone needs? It seems very very niche

validator/src/lib.rs
```diff
@@ -13,7 +13,15 @@ pub use validation::must_match::validate_must_match;
 pub use validation::non_control_character::validate_non_control_character;
 #[cfg(feature = "phone")]
 pub use validation::phone::validate_phone;
+pub use validation::range::validate_range_generic;
+
+#[deprecated(
+    since = "0.12.0",
+    note = "Please use the validate_range_generic function instead"
+)]
+#[allow(deprecated)]
```
Keats, line 22: I don't think anyone is using that crate without the derive, we can probably have the breaking change without deprecation.

validator/src/validation/range.rs
```diff
@@ -2,32 +2,45 @@ use crate::validation::Validator;
 
 /// Validates that a number is in the given range
 ///
-/// TODO: see if can be generic over the number type
+#[deprecated(since = "0.12.0", note = "Please use the validate_range_generic function instead")]
```
Keats, line 5: same, just change the code and we will bump the version

validator_derive/src/validation.rs
```diff
@@ -123,13 +123,13 @@ pub fn extract_range_validation(
                     match ident.to_string().as_ref() {
                         "message" | "code" => continue,
                         "min" => {
-                            min = match lit_to_float(lit) {
+                            min = match lit_to_float_or_path(lit) {
                                 Some(s) => Some(s),
                                 None => error(lit.span(), "invalid argument type for `min` of `range` validator: only integers are allowed")
```
Keats, line 128: we probably need to update the error messages if we allow float and not just integers...

Keats (changes requested):
Looks good but I'd remove the deprecation and just push the changes directly.

README.md
xFrednet: Yup this is very very specific or better say I have no real use case right now. 

The actual reason why this is here is the `ValueOrPath` enum. I think it would be beneficial to also use it for other things like the `Length` validation as well. I therefor added this functionality which was only four lines:
```rust
if path.starts_with("self.") {
    // Self value
    let ident = syn::Ident::new(path.trim_start_matches("self."), Span::call_site());
    quote!(self.#ident)
}
```

You can decide if we leave it in or not. It's also an easy change otherwise :)

validator/src/lib.rs
xFrednet: Alright, that makes some things easier :)

validator_derive/src/validation.rs
xFrednet: You're right, I missed that :).

validator/src/validation/range.rs
xFrednet: > and we will bump the version

I could also change the `length` validator to take a path as well before we push the next version. There is not issue for it but that should be simple enough and useful as well :) (But probably to much for this PR)

README.md
Keats: Let's remove it for now to keep the API small and easy to understand.

validator/src/validation/range.rs
Keats: Go for it in that PR if you want

xFrednet: I will :)

README.md
xFrednet: It has been removed :)

xFrednet:
I've the option to reference a value in the length validator as well :)

validator_derive/src/lit.rs
```diff
@@ -22,23 +23,58 @@ pub fn lit_to_float(lit: &syn::Lit) -> Option<f64> {
...
+
+    let path = lit_to_string(lit);
+    if let Some(path) = path {
+        return Some(ValueOrPath::Path(path));
+    }
+
+    None
+}
+
+pub fn lit_to_float_or_path(lit: &syn::Lit) -> Option<ValueOrPath<f64>> {
```
Keats, line 40: lit_to_f64 to be consistent with the other? Or the opposite

Keats (commented):
A small comment that can be done later. Thanks!

Issue #38, opened 2018-01-25
How to pass args to custom validators?


Boscop:
I want to have a validator that works like `length` but uses the byte length of a string (or rather, the length after converting utf8 to cp852).

How can I pass the len arg to that custom validation function?

Keats:
That's the most annoying thing about the custom validator for me right now as i don't really know how to do that. I'd love to be able to do a custom validator like `is_unique_username` that takes a db connection as first arg but I can't see how to pass it to the validate function.

I'd be very happy if someone can come up with a solution

Boscop:
That's actually exactly my other use case too, in my db I have `unique(customer_id, unique_id)` for all my tables so the `unique_id` only has to be unique **per customer** and to be able to validate, I need to pass the currently logged in customer's `customer_id` to the custom validator as context..

Keats:
For that specific case, it seems a `must_not_match` validator could be added as there is a `must_match`

csharad:
Yeah, that could be done for this particular use case. But I was outlining when we required a composite validation like the case above or comparing (field_a < field_b) or any other complex validation, an api could be provided so that we could construct it ourselves.

xFrednet:
I've been looking into how this could be implemented. This is a writeup of my current insights and a proposal how this could be implemented without breaking existing implementations. This focusses purely on the trait definitions for now:

## The problem
The ideal way to pass arguments to custom validation functions would probably be to call the `validate` function with extra arguments that would than be forwarded. Something like this:

```rust
// a, b, c are extra arguments
instance.validate(a, b, c);
```

The problem with this approach is that the current trait declaration doesn't support additional arguments:

```rust
pub trait Validate {
    fn validate(&self) -> Result<(), ValidationErrors>;
}
```

## The Proposal
One of the main focus points should be to implement this new functionality without breaking any existing code. This is my proposal on how we could do that:

1. We define a new trait that is used for validation with arguments like this:
    ```rust
    pub trait ValidateWithArgs<'a> {
        type Args;

        fn validate_args(&self, args: Self::Args) -> Result<(), ValidationErrors>;
    }
    ```
    The lifetime is used to pass arguments by reference. 
    
    This easily supports one or multiple arguments in the following why:
    ```rust
    // Single argument
    struct Duck;
    impl<'a> ValidateWithArgs<'a> for Duck {
        type Args = &'a Database;

        fn validate_args(&self, db: Self::Args) -> Result<(), ValidationErrors> {
            [...]
        }
[...]

Issue #96, opened 2020-03-08
#[NEW FEATURE] Attributes on function parameters


olvyko:
With release of [Rust 1.39.0](https://blog.rust-lang.org/2019/11/07/Rust-1.39.0.html#attributes-on-function-parameters) we have the opportunity to add attributes to function parameters.
I thought it would be cool to add feature like this:
```rust
#[macro_use]
extern crate validator_derive;
extern crate validator;

use validator::{Validate, ValidationError};

#[derive(Validate)]
struct Preference {
   #[validate(length(min = 4))]
   name: String,
   value: bool,
}

fn store_user(
    username: String,
    #[validate(email)] email: String,
    #[validate] preference: Preference,
) -> Result<(), ValidationError> {
    // 1. validations runs here
    // 2. function logic
    Ok(())
}
```

Keats:
What happens in 1? Is it inserting automatically some calls that the user won't see? Or do they still need to call a `validate` fn?
If it can be done wiht a nice UX/not too magical, I would take a PR for it

PR #137, opened 2020-11-30
Arguments for custom validation functions. #38

xFrednet:
This was quite an experience to create but a fun one :)

Okay, the implementation is pretty similar to the plan I described in #38. The implementation is based on the `ValidateArgs<'v_a>` trait:
```rust
pub trait ValidateArgs<'v_a> {
    type Args;

    fn validate_args(&self, args: Self::Args) -> Result<(), ValidationErrors>;
}
```

The usual `Validate` implementation simply references the `ValidateArgs<'v_a>` implementation. The custom validator now takes an `arg` parameter that can be set to one type. Here are some examples:
```rust
// fn validate_tuple(value: &str, tuple: (i64, i64)) -> Result<(), ValidationError>
#[validate(custom(function = "validate_tuple", arg = "(i64, i64)"))]
value: String
// instance.validate_args((77, 555));

// fn validate_with_db(value: &str, db: &mut Database) -> Result<(), ValidationError>
#[validate(custom(function = "validate_with_db", arg = "&'v_a mut Database"))]
value: String
// instance.validate_args(&mut db_instance);
```

More examples can be seen in the tests. The tests also include some error messages in the `compile-fail` folder.

---

The implementation also support arguments in multiple custom validators by packing them into a tuple like this:
```rust
#[derive(Debug, Validate)]
struct TestStruct {
    #[validate(custom(function = "validate_tuple", arg = "(i64, i64)"))]
    tuple_validation: String,

    #[validate(custom(function = "validate_with_db", arg = "&'v_a mut Database"))]
    db_validation: String
}
// Args type: ((i64, i64), &'v_a mut Database)
//
//                        |<-- Args tuple          -->|
// instance.validate_args(((77, 555), &mut db_instance));
//                         |<--1-->|  |<--    2   -->|
//      1. Tuple from tuple_validation
//      2. Reference from validate_with_db
```

This implementation doesn't support schema validation. But it should be extendable if that is required.

---

TODOs:
* [x] Improve error reporting for types with unknown size
* [x] Save `syn::Type` instead of string in the `Validator` enum

---

btw: I find the readme a bit unreadable due to it's length by now. I tried to use `<details>` tags to hide some content. Here is an example. I can merge it if this is interesting to you as well. It's sadly not the perfect solution -.- [Link](https://github.com/xFrednet/validator/blob/000-readme-change-suggestion/README.md)

<details>
<summary>Hello</summary>
Hello World
</details>


Keats:
That's impressive!
I don't know how I feel about the UX of it though, has anyone got some feedback?

xFrednet:
Thank you! I like the interface if you just have a single custom validation function and I guess that that covers 90% of the use cases. Packing a tuple for multiple functions is not ideal but it makes the system way more flexible. The only other interface that I can think of to support this would be to create a struct that holds the arguments. This if nice if you have multiple arguments but would be a bunch of overhead if you just want to use a single custom validation function.

This method is also fairly readable for multiple arguments if you store the data in values first. Something like this:
```rust
#[derive(Validate)]
struct ToBeValidated {
    // fn validate_tuple(value: &str, tuple: (i64, i64)) -> Result<(), ValidationError>
    #[validate(custom(function = "validate_tuple", arg = "(i64, i64)"))]
    one: String,
    
    // fn validate_with_db(value: &str, db: &mut Database) -> Result<(), ValidationError>
    #[validate(custom(function = "validate_with_db", arg = "&'v_a mut Database"))]
    two: String,

    // fn validate_with_db(value: &str, db: &mut Database) -> Result<(), ValidationError>
    #[validate(custom(function = "validate_with_db", arg = "usize"))]
    three: String,
}

let one_arg = (16, 21);
let two_arg = &mut db_instance;
let three_arg = 23;
instance.validate_args((one_arg, two_arg, three_arg));
```

omid:
@Keats How could we push it forward?

Keats:
I think some feedback on how it works for people, eg they try to use it in their codebase with cargo pointing to the git branch. I like it overall myself

omid:
@Keats Perfect, I'll try to put it in a production project I have and will report the potential issues/requirements.

omid:
@xFrednet I've found several issues in more complex situations.
And one case which does work in the latest version, but not in this branch.

So I've made a test project. It's a clone of an actix example, then I tried to add more complex models there.

Could you please clone and check it: https://github.com/omid/test-validator

One additional situation which you cannot see in that code above, if I forgot to add `'v_a`, the error message is like this (Note the `V<'v_a>alidate`):
```
error[E0106]: missing lifetime specifier
  --> src/models.rs:13:23
   |
13 | #[derive(Deserialize, Validate)]
   |                       ^ expected named lifetime parameter
   |
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider using the `'v_a` lifetime
   |
13 | #[derive(Deserialize, V<'v_a>alidate)]
   |                       ^^^^^^^
```

xFrednet:
Thank you very much for the effort @omid this is a major contribution! I'll try to check out your test project today, but it might take some time to get all the problems fixed as I'm also busy at work and in another project. Doing the work of testing is awesome and more than just appreciated!

A side node: The custom arguments are currently only working for the custom validator. Adding arguments to the schema validation should be simple but is something I wanted to do after this got accepted and merged :upside_down_face:

xFrednet:
> One additional situation which you cannot see in that code above, if I forgot to add `'v_a`, the error message is like this (Note the `V<'v_a>alidate`):

That error message is really not as helpful that's a good note. The only way I see to fix this is to check if a string contains a reference without a `'v_a` lifetime, and we then need to error on this. This could potentially reject valid parameters but would probably be worth it for the improved diagnostics. Especially since this is a more advanced feature :upside_down_face:

xFrednet:
Your test has already presented one bug where I know how to fix it :upside_down_face: Two comments about the usage: 

1. The `arg` parameter only takes the argument type. You often put the entire function signature with the result.
    ```rust
    // This
    #[validate(custom(
        function = "validate_email",
        arg = "(&'v_a diesel::pg::PgConnection, FnOnce(&str, &diesel::pg::PgConnection) -> Result<Option<User>, diesel::result::Error>)"
    ))]
    pub email: String,

    // Should be written as this:
    #[validate(custom(
        function = "validate_email",
        arg = "(&'v_a diesel::pg::PgConnection, FnOnce(&str, &diesel::pg::PgConnection))"
    ))]
    pub email: String,
    ```

    This is still broken due to some other error I'll look at :)
2. Nested validation sadly doesn't work with arguments. I believe that this is currently not possible with the control we with a macro crate. The only solution would be to add the arg parameter to the `#[validate]` part but this will get very chaotic. Restricting the arg validation to one layer seems better to me.

omid:
@xFrednet I tried to apply it to a real-world application :D
Maybe we need to think of a way to fix it.

Keats:
Lifetimes/sync/send are the potential issues I had in mind

xFrednet:
I'll look into getting your test project running, but it may take a bit of time due to my schedule. I'm sorry about that.

The nested validation would definitely be nice but probably hard to implement. The most beautiful solution would be to check what parameters are required by the nested object. But I believe that this is not possible due to the compiler stage we are working in and as a principle from Rust. I believe that macros were created in a way that they don't interact which each other, this is awesome but limits us in this case. (I can be wrong about this last part though). This means that we need to supply the nested parameter in some other way. Do you have any syntax suggestions for this?

I would push nested parameters and schema validation into another branch to not mix up the implementation too much. Having this after the bug fixes as a working version in a next branch seams better than adding more and more featured to this branch.

---

> Lifetimes/sync/send are the potential issues I had in mind

* Lifetimes, this is definitely a complicated aspect of this PR. I hoped that `' v_a` would cover everything, but it doesn't seam like it. I might look into changing that a bit since I'm now more used to work with the AST.
* sync and send shouldn't be a problem as we're not doing any multi-threading. There could be some difficulty with sync/send in the linked functions but that should be covered by the compiler. (This is just my theoretical view though) @omid is that being tested as a part of your project? I saw Diesel but I'm not sure about sync/send usage.

omid:
@xFrednet sorry, I'm not deep into the code and changes you've made. So I have no specific suggestion now. But will try to dive deeper this week, maybe find some ideas :-/

xFrednet:
Macros are pretty flexible, and we can define our own syntax when it comes to parameters. The limitation that my implementation currently has is that it can only store one type per validation entry meaning this:

```rust
//         vvvvvv We have a lot of freedom what we can write in here
#[validate(custom)]
//         ^^^^^^ This is what I mean with > "per validation entry"
```

Here are two suggestions that would be possible but I'm not to happy with them. They are not as readable and seam kind of inconsistent with the rest of the crate if you ask me:

```rust
// Option 1
#[validate(nested(arg="..."))]
// Option 2
#[validate(arg="...")]
```

---

I'm happy that you actually want to dig into the code. Feel free to ask me any time if something is unclear @omid. :upside_down_face:

omid:
@xFrednet @Keats
I have another suggestion (RFC), I'm not sure it will work or not.

Assumptions:
1. All extra arguments to a custom validation function should be immutable. An immutable reference can be enough! No?
2. Just custom validation functions need extra params, standard ones will never need!

Suggestion:
1. Instead of having arguments in the Struct, we can have them on the function side, similar to this:
```
#[validate]
pub fn validate_email(email: &str) -> R {
    // something
}
```
which will convert it to something like (I will talk about `Validate` later):
```
pub fn validate_email(email: &str, validate: &Validate) -> R {
    // ...
    let inner = fn inner(email: &str) -> R {
        // something
    }
}
```
2. We can/should name the input variables:
```
#[validate(db_connection)]
pub fn validate_email(email: &str, db: &DbConnection) -> R {
    // something
}
```
which will convert it to something like:
```
pub fn validate_email(email: &str, validate: &Validate) -> R {
    // Check Validate object, and return proper compile time errors
    let inner = fn inner(email: &str, db: validate.get("db_connection") -> R {
        // something
    }
}
```
3. Define the struct like before (master branch):
```
#[derive(Deserialize, Validate)]
pub struct NewUser {
    pub name: String,
    #[validate(custom(function = "validate_email"))]
    pub email: String,
    #[validate]
    pub numbers: Numbers,
}
```
4. But, in cases we have a custom function with custom parameters, we need to change the way we call validate, something like this:
```
new_user
    .validate_arg("db_connection", conn.deref())
    .validate();
```

Notes:
1. Most of the cases I can imagine needs database connection, so one argument definition can be enough for most cases:
```
#[derive(Deserialize, Validate)]
pub struct NewUser {
    #[validate(custom(function = "validate_username"))]
    pub username: String,
    #[validate(custom(function = "validate_email"))]
    pub email: String,
}

new_user
    .validate_arg("db_connection", conn.deref())
    .validate();
```
2. As far as I can see, it will work with nested validations.
3. I'm not sure if it can work with `Fn`s, like [this one](https://github.com/omid/test-validator/blob/master/src/validate_email.rs).

xFrednet:
I have to think about this for a bit these are just my initial thoughts in case you already want to add something to them:

Your assumptions are correct but here are two notes about them:
> 1. All extra arguments to a custom validation function should be immutable. An immutable reference can be enough! No?
    
* Having mutable references would be nicer to not limit the user and do enable validation that maybe needs to modify some metadata in the parameters. This problem can however be managed by using a [std::cell](https://doc.rust-lang.org/std/cell/) or some other smart pointer

> 2. Just custom validation functions need extra params, standard ones will never need!
    
* This is currently true and will probably remain to be true except for schema validation which is pretty much the same as custom validation

---

I'll sum up my understanding from your suggestion to make sure that I got it right:

1. We create a `Validate` struct which can basically map from an argument key like `"db_connection"` to the actual object reference
2. We create a wrapper around validation functions that takes the arguments out of the `Validate` struct instance and passes them to the function
3. Calling the validation first requires the loading of arguments for example with `.validate_arg("key", value)`

I see several problems with this:
1. This might be a bit of nagging, but it doesn't seem like the Rust way. Derive macros should usually only implement a trait like our `Validate` and in this PR `ValidateArgs`. This sounds more like creating a complete custom implementation.
2. In regard to separate function and struct macros. Each macro is parsed individually. The derive macro has access to the set metadata of the struct fields, but we won't be able to validate that a function which takes an email and a `Validate` instance exists. This is not a problem per se (Rocket etc. kind of work like this), but it can make it difficult to give helpful error messages.
3. Using the `#[validate]` on a function could have an effect regarding the syntax of #96 if it will ever be implemented
4. How can we make sure that each argument is actually set? 
    ```rust

    // Call
        new_user
            //.validate_arg("db_connection", conn.deref())
            .validate();

    // Arg request:
        //                                    vvvvvvvvvvvvvvvvvvvvvvvvvvvvv What happens now?
        let inner = fn inner(email: &str, db: validate.get("db_connection") -> R {
            // something
        }
    ```

    This is more an implementation detail which can easily be fixed by wrapping each function argument in `Option<XYZ>` but it's still something to think about.
5. About your note:
    > 2. As far as I can see, it will work with nested validations.

    This will not happen directly as we simply call `struct.field.validate()` which wouldn't automatically have access to the arguments of this validation. This should be doable by simply passing the arguments on to the struct though.

I'm sorry that this might sound negative as I don't think that this is the right way. There was already some discussing in the issue itself: #38. A high focus of mine is to not break any existing implementations and not to change the usage too much. However, you did remind me of an idea that was listed in these examples [here](https://github.com/Keats/validator/issues/38#issuecomment-733818713) which could be a possibility. We could use one struct like a builder struct to save the arguments in between. IDK if this would solve our problems, but it could be as well as help which readability? This as well as your suggestion is definitely something I'll think about this weekend. 

Thank you for giving this some though @omid. It definitely helps and I highly appreciate it! :)

Keats:
I agree with @xFrednet, I think this solution would be worse in practice than this PR but it was an interesting idea

omid:
Thanks for the feedback :)

@xFrednet Yep, you got it correctly.

I agree, my suggestion can face more runtime errors/panics than the derive way.

About the problems you mentioned
1. I agree that it will be not just a derive macro and will be a derive and attribute macros. Is there something logically against it? (I'm asking to learn :D )
3. I cannot see the problem, simply we can rename this or that to something else, like `#[validate_param]` for example.
5. Internally we may need to pass the `Validate` struct everywhere.

xFrednet:
About your questions @omid:
> 1. I agree that it will be not just a derive macro and will be a derive and attribute macros. Is there something logically against it? (I'm asking to learn :D )

- There is nothing logical speaking against this. As said it would most likely also work. It's more a thing of expectations and simplicity. A derived implementation should be simple and self contained (if possible). Requiring the user so also add a attribute to the function would go a bit against the self contained thing.  This is nothing purely logical though.

> 2. I cannot see the problem, simply we can rename this or that to something else, like `#[validate_param]` for example.

- Yes of course this is true. This was more a thing to give it some thought. Why limit possible future usage or syntax if we implement something that is not completely related to the function?
- This actually made me think of a second point. Using the `#[validate]` attribute would require a wrapper for functions that are in external crate. A direct call would not require this.

> 3. Internally we may need to pass the Validate struct everywhere.

- Exactly, this is a probability, however it might be avoidable to make the code generation easier and to avoid the overhead even if it is pretty little.

Did this clear some things up? :upside_down_face:

xFrednet:
Status update:

I'm looking into the test project now. A part of the reports are the result of objects were the size is not known at compile time. The reports a pretty much the same as it would be without the derive implementation. We therefor only have to improve the actual error reporting. I have a plan how this can be done using [`quote_spanned!`](https://docs.rs/quote/1.0.8/quote/macro.quote_spanned.html).

<details>
<summary>Here is a simple pure rust example which produces the same reports about unknown size of type at compile time:</summary>

### main.rs
```rust
fn test_me(a: FnOnce(&str) -> usize) {
    println!("a: {}", a("HELLO"));
}

fn main() {
    test_me(|x| x.len());

    println!("Hello World!");
}
```

### cargo check output
```
warning: trait objects without an explicit `dyn` are deprecated
 --> src/main.rs:1:15
  |
1 | fn test_me(a: FnOnce(&str) -> usize) {
  |               ^^^^^^^^^^^^^^^^^^^^^ help: use `dyn`: `dyn FnOnce(&str) -> usize`
  |
  = note: `#[warn(bare_trait_objects)]` on by default

error[E0277]: the size for values of type `(dyn for<'r> FnOnce(&'r str) -> usize + 'static)` cannot be known at compilation time
 --> src/main.rs:1:12
  |
1 | fn test_me(a: FnOnce(&str) -> usize) {
  |            ^ doesn't have a size known at compile-time
  |
  = help: the trait `Sized` is not implemented for `(dyn for<'r> FnOnce(&'r str) -> usize + 'static)`
help: function arguments must have a statically known size, borrowed types always have a known size
  |
1 | fn test_me(&a: FnOnce(&str) -> usize) {
  |            ^

error[E0308]: mismatched types
 --> src/main.rs:6:13
  |
6 |     test_me(|x| x.len());
  |             ^^^^^^^^^^^ expected trait object `dyn FnOnce`, found closure
  |
  = note: expected trait object `(dyn for<'r> FnOnce(&'r str) -> usize + 'static)`
                  found closure `[closure@src/main.rs:6:13: 6:24]`

error[E0277]: the size for values of type `(dyn for<'r> FnOnce(&'r str) -> usize + 'static)` cannot be known at compilation time
 --> src/main.rs:6:5
  |
6 |     test_me(|x| x.len());
  |     ^^^^^^^ doesn't have a size known at compile-time
  |
  = help: the trait `Sized` is not implemented for `(dyn for<'r> FnOnce(&'r str) -> usize + 'static)`
  = note: all function arguments must have a statically known size

error: aborting due to 3 previous errors; 1 warning emitted

Some errors have detailed explanations: E0277, E0308.
For more information about an error, try `rustc --explain E0277`.
```

</details>

xFrednet:
Working on the size fix actually gave me an idea for a cleaner call syntax. We could use a macro to have a slightly better calling syntax:

```rust
// Macro without args
    validate!(object);
    // equivalent to
    object.validate();

// Macro with args
    validate!(object, arg1, arg2, arg3);
    // equivalent to
    object.validate_args((arg1, arg2. agr3));
```

The user doesn't even notice the tuple packing hackery and has a simple syntax for validation with and without arguments. What do you think about it? :upside_down_face:

xFrednet:
Okay, I've improved the error message for invalid argument types by restricting them to only types were we know for sure that they are valid this would be:
* Tuples if all contained types are valid
* Direct number types, as they will be copied
*  Types which are referenced by `'v_a`

All other types like `String` or `FnOnce(&str) -> usize` will now receive an error message stating that they should use a reference.

---

This means that the test project is still broken, but the error message should be much more helpful and straight forward to fix. @omid Could you maybe look into fixing the project as a test to see if the error messages are clear enough?

---

This still leaves open the request from @omid to allow custom arguments for nested validation. This can be implemented but as I mentioned here ([comment](https://github.com/Keats/validator/pull/137#issuecomment-767041595)) I need to have a defined syntax for it. Some input would be greatly appreciated otherwise I can just go with the `#[validate(nested(arg="..."))]` option.

Keats:
Had another look tonight and it looks good to me. I'm not sure about nested validation yet, maybe we could release this and get more feedback?

xFrednet:
I would like to add custom arguments to the `schema` validation before releasing the next version, but it would be good to get this already on the next branch and add the `schema` validation in a new PR. I'll resolve the merge conflicts in the weekend and try to get into the schema validation that might take a few weeks though, due to some other work :upside_down_face:

xFrednet:
I've updated from master and this branch is ready for review/merge into the next branch. The next step will then be to add the schema validation and that should hopefully be it :upside_down_face:

dyedgreen:
As some prior art to this: The Juniper GraphQL server framework has a notion of `Context`, and anything which can be resolved has optionally access to this context: https://docs.rs/juniper/0.15.4/juniper/executor/trait.Context.html

One interesting aspect about how the context works, is that they can be converted into each other, e.g. the default context is unit, and can be converted to from any context. So you can mix and match them.

I haven't looked into the macros of this crate too heavily, but I believe something like this would work (going roughly from the juniper example):

```rust
trait Context {}

impl Context for () {}

trait Validate {
  type Context: Context;

  fn validate<C>(&self, ctx: &C) -> Result<(), ValidationErrors>
  where
    Self::Context: FromContext<C>;
}

trait FromContext<T: Context> {
  fn from(other: &T) -> &Self;
}

impl<T: Context> FromContext<T> for () {
  fn from(_: &T) -> &Self {
    &()
  }
}
```

(The approach taken in juniper is simpler, since they only allow passing the context by reference; but in my experience this is usually not a problem: If we have something like a DB connection, it's enough to have a reference. If we want to have a short-lived local state, we can go with a `Cell`, especially since this is not `async` / threaded in any way.)

videni:
can't wait for this feature.

Keats:
@xFrednet what do you think of the Juniper approach?

xFrednet:
I'm not quite sure that I fully understand the idea. From the examples it seems like we have a `context` type than is turned into each of the required argument.

Based on this assumption, I would say that the context interface seems cleaner, but also has a bigger overhead. This PR implementation simply requires the definition of arguments, and you're good to go, while the context implementation requires a user made impl block for each different argument. It's likely that context implementations could also be reused, if the user often passes the same arguments.

So, based on all of this, I'm not sure. I have a bias towards this implementation due to the time spend on it, but that's it. The best step would probably be to implement a prototype using the context way and then compare the two. It would be great if someone could do that :upside_down_face:

Keats:
I think I prefer your implementation as well since it's less setup for the user

xFrednet:
Okay, in that case, do we want to move this implementation to the `next` branch? As there haven't been any statements against this.

 I still would like to add the arguments to the `schema` validation as well, this will also leave some time for further discussion :upside_down_face:. I'm currently stuck in another project, but I should hopefully find some time to add that in the next two weeks, unless someone else would like to do that.

Keats:
@xFrednet I think we can release it as the next version directly?

svenstaro:
@Keats @xFrednet As an interested third-party I'm very much looking forward to this and would certainly appreciate this getting merged and released. How about we put it on `next` now and follow up on further improvements there?

raffomania:
Two questions that came to mind when reading the source:

- If I have a single struct with two different custom validation functions that need different parameters, e.g. `transaction` and `http_client`, I'm assuming I would need to pass both parameters to both functions inside a tuple: `(transaction, http_client)`. Is that correct?
- Is the `'v_a` lifetime equivalent to the lifetime of the validator? If so, does this place any restrictions on the kinds of lifetimes that are allowed for references in `args`?

I'm curious to dive deeper into this and will try the branch out in our project, thank you a lot for putting in all the work!

xFrednet:
@Keats We can release it as is, but I think it would feel a bit unfinished without the `schema` validation having custom arguments as well. That's the main reason why I would not release it right away.

@svenstaro Thank you for the support. We're currently discussing if this will need any further improvements before releasing :upside_down_face: 

@raffomania your first assumption is correct, note that you'll also need to pass them as a tuple if they would be the same argument.
```rust
#[validate(custom(function = "a", arg = "i64"))]
#[validate(custom(function = "a", arg = "i64"))]
```

Would require you to call `instance.validate_args((value_1, value_1))`

The lifetime `'v_a` is required to pass pointers to functions. It is only used by the arguments when calling `validate_args` and will release everything right away. Here is an example

```rust
println!("Some awesome code");

// We call the validation here. The life of 'v_a starts here
instance.validate_args((value_1, value_2))
// and 'v_a is ended here. 
// (Unless you do some funky stuff of storing references with the 'v_a lieftime inside some global value (please dont))

println!("Some awesome code");
```

Does this example make it a bit more clear? Feel free to reach out if you have any questions :upside_down_face:

raffomania:
Thanks for the detailed explanation! Thinking about the lifetime, it seems like it shouldn't cause too many difficulties :)

One further thing I noticed: Some people in this thread have already mentioned using database transactions as arguments, and in our project we would like to do this as well. However, with databases it's possible to encounter random failures such as `std::io::Error` due to the database being unavailable, network failure etc.

AFAIK, there's currently no possibility to handle this case well. `validator` assumes that validation cannot fail. Of course, it's possible to panic, but that's not feasible in some situations. So for using this feature in conjunction with databases, I think we'd need a second PR with new error-handling capabilities :)

Keats:
Argh, interesting I didn't think of errors

xFrednet:
I think it would be okay to require custom validation functions to return a different type like an extra `CustomValidationError` as it wouldn't break any existing code. But I'm worried that this will also make the API a bit more complicated and inconsistent. Introducing a new return type could also cause more problems in the long run.

The current implementation offers the opportunity to also pass `mut` arguments that could also be used to pass errors along, but that seems very hacky and not rust like. Still, here is some pseudocode: 
```rust
#[validate(custom(function = "validate_tuple", arg = "(&'v_a Database, &'v_a mut Result<(), ()>)"))]

fn validate_tuple(value: &str, (db, mut result): (&Database, &mut Result<(), ()>)){
    if db.execute().is_err() {
        result = Err(());
    }
}
```

Could you give an example what you would like to do with the error that would occur in the validation? You can easily pass a message to the user via the current framework. The main reason to add a new type would be to add some special handling that can only be done outside the validator, correct? :upside_down_face:

raffomania:
The use case I'm thinking of is that in a web context, validation errors should result in a 422 HTTP response status code, while database connectivity errors should result in a 500 status code. Of course, it would be possible to create a validation error with a special error code or some similar mark, but that would make it easier to introduce mistakes, i.e. when mistyping the code string or forgetting to transform the error after calling the validator.

Keats:
I think the approach with `CustomValidationError` makes sense, I expect most of the custom functions will want to use their own error but at the same time, we need to consider how easy it is to use afterwards. @dyedgreen can you try to implement it just to see how it would work in practice?

Keats:
Let's release as is for now and we can improve on a next version.

xFrednet:
Okay, do you need any further help with the release?

Keats:
I'll merge everything and see what happens

PR #140, opened 2020-12-28
add domain validation

zhangmingkai4315:
validator crate has a private method ```validate_domain_part()``` in email.rs,  this method maybe enough for email validation, but as far as i know,  it not suitable for all domain names, like [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) and [SRV](https://en.wikipedia.org/wiki/SRV_record) type domain name, such as ```_jabber._tcp.gmail.com``` ,  so i think it may be useful to write a independent method for  domain validate only.  

There are some RFC for domain name definition and modification, such as  [RFC 1034](https://tools.ietf.org/html/rfc1034),[RFC 1035](https://tools.ietf.org/html/rfc1035) and  [RFC 2181](https://tools.ietf.org/html/rfc2181).

### DIFF
--- a/validator/src/lib.rs
+++ b/validator/src/lib.rs
@@ -71,6 +71,7 @@ mod validation;
 pub use validation::cards::validate_credit_card;
 pub use validation::contains::validate_contains;
 pub use validation::email::validate_email;
+pub use validation::domain::validate_domain;
 pub use validation::ip::{validate_ip, validate_ip_v4, validate_ip_v6};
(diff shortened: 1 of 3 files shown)


Keats:
The current domain validation does miss some valid domains since it follows the HTML spec which doesn't cover 100% of valid emails. I don't think I want to add more though, or maybe just fix the existing regex to fit more cases.

PR #145, opened 2021-02-14
Changed EMAIL_DOMAIN_RE'gex

Belanchuk:
Validator skips invalid emails addresses like: name@domain.
I changed the relevant regex and the test.


Keats:
It can be valid in some cases, eg `abc@localhost`. https://stackoverflow.com/questions/5601647/html5-email-input-pattern-attribute.
I would prefer to keep the HTML regex so it can be consistent with the (potential) frontend.

Belanchuk:
> It can be valid in some cases, eg `abc@localhost`. https://stackoverflow.com/questions/5601647/html5-email-input-pattern-attribute.
> I would prefer to keep the HTML regex so it can be consistent with the (potential) frontend.

Is localhost the only valid case in which the domain part of the address consists of one part?
Maybe add localhost to the regular expression?
```rust
("abc@bar", false),
("abc@localhost", true),
```

PR #147, opened 2021-03-09
Simplify validate_ip_v{4,6} methods

buinauskas:
Instead of matching, clippy suggests using `is_ipv4()` or `is_ipv6()` methods. 

Applied these, also refactored to simply use `.map_or(...)` instead of matching.


Keats (approved):

Keats:
Thanks!

PR #154, opened 2021-05-02
Enhancement/support u128 and i128 and their Option variants

jsj1027:
Hi, so I looking at some of the issues and decided to contribute this.

These changes update the code and tests to include u128 and i128 support along with their Option variants.

I did run into something I couldnt solve though.

In my own testing this panics but not for the reason I expected
``` rust
use serde::Deserialize;

// A trait that the Validate derive will impl
use validator::{Validate, ValidationError};

#[derive(Debug, Validate, Deserialize)]
struct SignupData {
    #[validate(range(min = 18, max = 20))]
    age: u128,
}

fn main() {
    println!("Hello, world!");
    let hi = SignupData{ age: 17u128};
    println!("Hi you are {:#?} years old", hi.age); 
    hi.validate().unwrap();
}
```

The error message state:
``` rust
Hello, world!
Hi you are 17 years old
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("u128 is not supported", line: 0, column: 0)', /home/kaloshade/Code/validator/validator/src/types.rs:22:48
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

However the same code but with the age value set within the range, works just fine. The same thing happens with i128 Option<u128> and Option<i128>,
Another interesting note is that all the tests btw the libraries still pass weirdly enough,

I attempted to figure this out for a day or two with no luck, so I figured I'd create a PR and maybe someone more knowledgeable would be able to buff out that issue.

### DIFF
--- a/validator/Cargo.toml
+++ b/validator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "validator"
-version = "0.13.0"
+version = "0.13.1"
 authors = ["Vincent Prouillet <hello@vincentprouillet.com"]
 license = "MIT"
 description = "Common validation functions (email, url, length, ...) and trait - to be used with `validator_derive`"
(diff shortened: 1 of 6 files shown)


GREsau:
That error occurs because Validator sets the "value" param on the generated error to the value which failed validation, in this case `17u128`. To do this, it converts the value to a `serde_json::Value` - which unfortunately does not accept u128/i128, so it falls back to the default serialization behaviour of returning a "not supported" error.

Keats:
Yeah.. I'm not sure what to do about that. Any ideas?

Issue #80, opened 2019-09-24
Easy way to get a Vec<ValidationError>


jamessewell:
I'm using nested validations, and it would be great if there was a simple way to get at the ValidationError objects.

At the moment it's obscured by the many levels of ValidationErrors

Keats:
How would that work? A `ValidationError` doesn't have the name of the field the error is on so you would get a bunch of errors but not know where it actually comes from.

jamessewell:
True.

How about a human readable display for the whole thing?

jamessewell:
Ok so thinking about this more I would really love something like what `serde_json` does:

```
Json deserialize error: unknown field `orthing`, expected `easting` or `northing` at line 1 column 365
```

But I suppose you need proper serde hooks to get line and column info.

Maybe something like:

```
Validate error: field `easting`, value 181 is not between -180 and +180
```

Sure it doesn't include any of the nesting information, but it's good enough for the majority of cases.

incker:
You can get a field name, my code snipet:

        my_struct.validate().map_err(|err: ValidationErrors| {
            let errors: HashMap<&str, validator::ValidationErrorsKind> = err.into_errors();
            for (field, err) in errors {
                // do what you need
            }
        })

I it good thing if you will make own wrapper over validator

And by the way, @jamessewell, you can make custom message errors, like `#[validate(length(min = 1, max = 255, message = "your good readable error"))]`

+ close issue if I have answered all questions

Keats:
Not all errors have a key. For example what would be the param key for a failed validation of the second string in a `vec` of a nested key? It could be an `Option` but it wouldn't be that great either since it would only be there in some cases.

dyedgreen:
I was also running into this issue. It would really help in my case, if the error `Display` was more presentable, such that it could e.g. be echoed out directly to users. (Example use case would be returning this error from a GraphQL api, and just displaying the error on the client side.)

@Keats I was thinking about a format similar to this:

### Format individual errors
```rust
match (message, code) {
  (None, None) => /* some default format, e.g. value 181 is not between -180 and +180 */,
  (Some(msg), None) => msg.to_string(),
  (Some(msg), Some(code)) => format!("{} ({})", msg, code),
}
```

### Format `ValidationErrors` collection
- for each error on a field do `{field_name}: {error1}, {error2}, ..., {errorN}`
- for nested structs recursively do `{field_name}.{nested_errors}` (so that you get `nested.field.name: Some error here`)
[...]

PR #158, opened 2021-06-02
Better display implementation

dyedgreen:
As discussed a little in #80, this PR adds a slightly nicer `Display` implementation which prints the paths to each error, and then either a generic description of the error, or the provided error message.

See the added tests for the format.

### DIFF
--- a/validator/src/display_impl.rs
+++ b/validator/src/display_impl.rs
@@ -1,0 +1,73 @@
+use crate::{ValidationError, ValidationErrors, ValidationErrorsKind};
+use std::fmt::{self, Write};
+
+impl fmt::Display for ValidationError {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if let Some(msg) = self.message.as_ref() {
+            write!(fmt, "{}", msg)
+        } else {
+            write!(fmt, "Validation error: {} [{:?}]", self.code, self.params)
+        }
+    }
+}
+
+fn display_errors(
+    fmt: &mut fmt::Formatter<'_>,
+    errs: &ValidationErrorsKind,
+    path: &str,
+) -> fmt::Result {
+    fn display_struct(
+        fmt: &mut fmt::Formatter<'_>,
+        errs: &ValidationErrors,
+        path: &str,
+    ) -> fmt::Result {
+        let mut full_path = String::new();
+        write!(&mut full_path, "{}.", path)?;
+        let base_len = full_path.len();
+        for (path, err) in errs.errors() {
+            write!(&mut full_path, "{}", path)?;
+            display_errors(fmt, err, &full_path)?;
+            full_path.truncate(base_len);
+        }
+        Ok(())
+    }
+
+    match errs {
+        ValidationErrorsKind::Field(errs) => {
+            write!(fmt, "{}: ", path)?;
+            let len = errs.len();
+            for (idx, err) in errs.iter().enumerate() {
+                if idx + 1 == len {
+                    write!(fmt, "{}", err)?;
+                } else {
+                    write!(fmt, "{}, ", err)?;
+                }
+            }
+            Ok(())
+        }
+        ValidationErrorsKind::Struct(errs) => display_struct(fmt, errs, path),
+        ValidationErrorsKind::List(errs) => {
+            let mut full_path = String::new();
+            write!(&mut full_path, "{}", path)?;
+            let base_len = full_path.len();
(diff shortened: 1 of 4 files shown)


svenstaro:
This would be excellent!

Keats:
It looks like the tests fail?

dyedgreen:
Oh, I think the issue is the test uses a derived validator, but the `--no-default-features` test run does not support this, causing it to fail compiling. Let me fix that 😅

Keats:
Thanks, I'll release that with #137

svenstaro:
@Keats Any chance for a merge and release of this before #137? It'd be already very for us on its own.

dyedgreen:
If you need this very urgently @svenstaro, you could specify an [override](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#working-with-an-unpublished-minor-version) for the PRs branch, until the next version is published on crates.io 😄

Keats (approved):

azzamsa:
Thanks for releasing it. I need this feature in production. I don't need to point any git branch anymore.

Keats:
Be careful when using that in production though, it might leak PII / passwords.

azzamsa:
```rust
use validator::{Validate, ValidationError};
use yaserde_derive::YaDeserialize;

#[derive(Debug, YaDeserialize, Validate)]
pub struct Person {
    #[validate(length(min = 1), custom = "validate_username")]
    name: String,
    age: i32,
}

fn validate_username(username: &str) -> Result<(), ValidationError> {
    let first2: String = username[0..2].into();
    if first2 != "foo" {
        let msg = "Invalid username format, it should looks like `foo_...`";
        return Err(ValidationError::new(msg));
    }
    Ok(())
}

fn main() {
    let input = "<Person><name>bar_aladdin</name><age>20</age></Person>";
    let person: Person = yaserde::de::from_str(input).unwrap();
    match person.validate() {
        Ok(_) => println!("Person: {:?}", person),
        Err(err) => println!("error: {}", err),
    }
}
```

With the code above, I got:

```rust
error: ValidationErrors({"name": Field([ValidationError { code: "Invalid username format, it should looks like `foo_...`", message: None, params: {"value": String("bar_aladdin")} }])})
```
Is there a way to get only *"Invalid username format, it should looks like `foo_...`"* without the rest of the error message?

dyedgreen:
@azzamsa you can attach a custom `message="…"` which will be displayed as the error message. (Notice that message is None)

azzamsa:
Thanks. I thought both feature are conflicting. I can't use `message` if I use `custom`, turns out it is not.
 
Sorry, but how to access the `message`. I tried `err.message` but it doesn't work.

```rust
error: ValidationErrors({"name": Field([ValidationError { code: "Invalid username format, it should looks like `foo_...`", message: Some("Invalid username format, it should looks like `foo_...`"), params: {"value": String("bar_aladdin")} }])})
```

azzamsa:
I dig the code-base and read the [test](https://github.com/Keats/validator/blob/7e7ba65d7fef72b02fb9220c91abe8043ce37281/validator_derive_tests/tests/schema.rs#L66). I found a way to access the `message`.

```rust
match person.validate() {
        Ok(_) => println!("Person: {:?}", person),
        Err(err) => println!("error: {:#?}", err.field_errors()["name"][0].message),
    }
```
```rust
error: Some(
    "Invalid username format, it should looks like `foo_...`",
)
```

Is there any better approach to access `message` rather than `...["name"][0].message` ?

azzamsa:
It works now. I need to use `#[validate(custom(function = "validate_username", message = "message_str"))]` attribute style.

```rust
use validator::{Validate, ValidationError};
use yaserde_derive::YaDeserialize;

#[derive(Debug, YaDeserialize, Validate)]
pub struct Person {
    #[validate(custom(function = "validate_username", message = "message_str"))]
    name: String,
    age: i32,
}

fn validate_username(username: &str) -> Result<(), ValidationError> {
    let first2: String = username[0..2].into();
    if first2 != "foo" {
        let msg = "Invalid username format, it should looks like `foo_...`";
        return Err(ValidationError::new(msg));
    }
    Ok(())
}

fn main() {
    let input = "<Person><name>aaaa</name><age>20</age></Person>";
    let person: Person = yaserde::de::from_str(input).unwrap();
    match person.validate() {
        Ok(_) => println!("Person: {:?}", person),
        Err(e) => println!("error: {}", e),
    }
}
```
```rust

     Running `target/debug/playground`
error: name: message_str
```

PR #160, opened 2021-06-23
Fix parsing of schema validation function

chjdev:
A similar fix to: https://github.com/Keats/validator/blob/e39ec89f4026e97e3089dabc40e33129a1053a6b/validator_derive/src/quoting.rs#L376

This allows referencing schema validation functions from different modules like:
`#[validate(schema(function = "crate::some_crate::validate_something"))]`

closes: https://github.com/Keats/validator/issues/161

### DIFF
--- a/validator_derive/src/quoting.rs
+++ b/validator_derive/src/quoting.rs
@@ -495,7 +495,8 @@ pub fn quote_field_validation(
 }
 
 pub fn quote_schema_validation(v: &SchemaValidation) -> proc_macro2::TokenStream {
-    let fn_ident = syn::Ident::new(&v.function, Span::call_site());
+    let fn_ident: syn::Path = syn::parse_str(&v.function).unwrap();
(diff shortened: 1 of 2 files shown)


validator_derive/src/quoting.rs
```diff
@@ -495,7 +495,8 @@ pub fn quote_field_validation(
 }
 
 pub fn quote_schema_validation(v: &SchemaValidation) -> proc_macro2::TokenStream {
-    let fn_ident = syn::Ident::new(&v.function, Span::call_site());
+    let fn_ident: syn::Path = syn::parse_str(&v.function).unwrap();
+    //let fn_ident = syn::Ident::new(&v.function, Span::call_site());
```
Keats, line 499: can you add a test and remove the commented out line?

chjdev: oops, I've removed the line
also added tests

Keats (approved):

PR #169, opened 2021-09-06
Derive Deserialize for ValidationErrors

Person-93:
Needed to change keys to String because the Deserializer can't provide &'static str

### DIFF
--- a/validator/src/types.rs
+++ b/validator/src/types.rs
@@ -41,7 +41,7 @@ pub enum ValidationErrorsKind {
 }
 
 #[derive(Default, Debug, Serialize, Deserialize, Clone, PartialEq)]
-pub struct ValidationErrors(HashMap<String, ValidationErrorsKind>);
+pub struct ValidationErrors(HashMap<Cow<'static, str>, ValidationErrorsKind>);
 
 impl ValidationErrors {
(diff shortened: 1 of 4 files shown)


Keats:
What is the usecase there? I'm expecting Validator to only ever serialize things

Person-93:
@Keats I need it so that validation can be done on the server and error handling can be done on the client.

Person-93:
I'm a bit confused by this error, is it a dependency that doesn't compile on the pinned toolchain?

Keats:
> I'm a bit confused by this error, is it a dependency that doesn't compile on the pinned toolchain?

Person-93:
Yes, it should work.
I was concerned that deserializing a `Cow<str>` would give it the lifetime of wherever it's derserialized from, but based on [this issue](https://github.com/serde-rs/serde/issues/1852) it looks like it does not and there are no plans to change that.
I'll fix it.

Person-93:
> Can you rebase your PR?

I've deleted my fork, but I looked around and found [this commit](https://github.com/Keats/validator/tree/e0c9856a49ccf125b22d0fecba7d51ea9a4743aa) which looks like it has all my changes. Feel free to clone it and open your own PR.

PR #174, opened 2021-11-09
Update README

apiraino:
Hi!

I think after the latest code refactor the README has now two dead links. I've tried to update them :-)

Thanks for checking them.

And thanks for this crate!


Keats (approved):
Thanks!

Issue #176, opened 2021-11-24
Allow validation of "string-like" types


cortopy:
I have code for protobuf messages generate with prost, which will be something like

```rust
struct Address {
  line1: prost::alloc::string::String
}
```

for most purposes `line1` is just a `String` and any function that works with one will work with the other.

However, because of the following function https://github.com/Keats/validator/blob/7e7ba65d7fef72b02fb9220c91abe8043ce37281/validator_derive/src/asserts.rs#L72

I can't use the derive annotations

Would it be acceptable to write the check as a match for "String"?

loyd:
+ also for `CompactString` and other strings with SSO

Keats:
This would be fixed by https://github.com/Keats/validator/issues/205

PR #177, opened 2021-11-24
allow any string-like value

cortopy:
As per https://github.com/Keats/validator/issues/176, I've tried to relax the validation a bit. It does work for prost Strings. Would this be acceptable?

### DIFF
--- a/validator_derive/src/asserts.rs
+++ b/validator_derive/src/asserts.rs
@@ -70,14 +70,7 @@ pub static NUMBER_TYPES: [&str; 38] = [
 ];
 
 pub fn assert_string_type(name: &str, type_name: &str, field_type: &syn::Type) {
-    if type_name != "String"
-        && type_name != "&str"
-        && !COW_TYPE.is_match(type_name)
-        && type_name != "Option<String>"
-        && type_name != "Option<Option<String>>"
-        && !(type_name.starts_with("Option<") && type_name.ends_with("str>"))
-        && !(type_name.starts_with("Option<Option<") && type_name.ends_with("str>>"))
-    {
+    if !type_name.contains("String") && !type_name.contains("str") {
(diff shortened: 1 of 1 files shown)


Keats:
It's not really going to cover your usecase if you have an Option for example. Maybe we just look if `str` or `String` are in it?

cortopy:
@Keats you were totally right. As I carried on adding validation I stumbled upon more cases.

I've added changes for `Option<*String>` and it's working for me

While doing this I realised that the main issue is that prost supports `no_std` and this impacts both String and Option (prost uses `core::option::Option`), but the latter is a different story altogether I think

I hope this PR helps with widening compatibility

validator_derive/src/asserts.rs
```diff
@@ -113,7 +113,7 @@ pub fn assert_has_len(field_name: String, type_name: &str, field_type: &syn::Typ
         return;
     }
 
-    if type_name != "String"
+    if !type_name.ends_with("String")
         && !type_name.starts_with("Vec<")
         && !type_name.starts_with("Option<Vec<")
         && !type_name.starts_with("Option<Option<Vec<")
```
Keats, line 119: should we just search for String and str rather than this ever-expanding list of possibilities? Worst case, it will be a compile time error if they tried to use it on a type it can't be used but that can be documented in the README.

cortopy: it does sound like a sensible approach to me. The validation error feels like a very nice message for easier development, but if it's getting too much on the way then relaxing it would help too.

I've made the change to simplify this as much as possible. There are still `Option<Option<` checks for other types but they don't affect the case of strings

Keats: We should keep the same logic for the length validator below though

PR #178, opened 2021-12-20
Implement `HasLen` for map/set types

nu11ptr:
This PR implements `HasLen` on all std::collection Map/Set types along with an optional feature to support the popular `indexmap` crate.

### DIFF
--- a/validator/Cargo.toml
+++ b/validator/Cargo.toml
@@ -22,6 +22,7 @@ validator_derive = { version = "0.14", path = "../validator_derive", optional = 
 card-validate = { version = "2.2", optional = true }
 phonenumber = { version = "0.3", optional = true }
 unic-ucd-common = { version = "0.9", optional = true }
+indexmap = {version = "1", features = ["serde-1"], optional = true }
 
 
 [features]
--- a/validator/src/traits.rs
+++ b/validator/src/traits.rs
@@ -1,5 +1,8 @@
 use std::borrow::Cow;
-use std::collections::HashMap;
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
+
+#[cfg(feature = "indexmap")]
+use indexmap::{IndexMap, IndexSet};
 
 use crate::types::ValidationErrors;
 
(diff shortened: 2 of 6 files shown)


validator/Cargo.toml
```diff
@@ -22,6 +22,7 @@ validator_derive = { version = "0.14", path = "../validator_derive", optional =
 card-validate = { version = "2.2", optional = true }
 phonenumber = { version = "0.3", optional = true }
 unic-ucd-common = { version = "0.9", optional = true }
+indexmap = {version = "1", features = ["serde-1"], optional = true }
```
Keats, line 25: why the serde feature?

nu11ptr: Essentially this allows you to use Serde for indexmap - without it your whole struct could be serializable but the indexmap wouldn't be (which of course wouldn't work)

Keats: But we are implementing the trait on the base types no? I don't know the indexmap crate but it feels like we shouldn't need to use that feature

nu11ptr: The trait is implemented on the collection types AND indexmap/set, a very popular crate (ordered hash map/set - filling a stdlib gap). Indexmap code is only included if the feature is specified of course, so users don't pay the price for it unless they want it. If I didn't include the `serde-1` feature than the users wouldn't be able to serailize/deserailize any structs with an indexmap. I included the feature so that it would behave as expected to users (since you include `serde` - it is not a new dependency).

Keats: Ah it's for passing the value to the error struct, I forgot about that requirement

jgrund:
Possible to refresh this patch and merge? In the current form we are stuck using Vec for collections or writing our own validator for checking length.

nu11ptr:
While we wait, you are welcome to use my fork as I am. Cargo makes it a breeze. In `[dependencies]`, just swap out your existing line to be something like this (assuming you want `indexmap` support):

```toml
validator = { git = "https://github.com/nu11ptr/validator", features = ["derive", "indexmap"] }
```

Keats:
Yes I'll merge it sometimes this week. I just need to refresh a bit on this repo, haven't looked at it directly in ages and there are a few other PRs

PR #179, opened 2021-12-24
Correcting clippy warnings

amoriarty:
Hello here !

First of all, since it's my first message on this repository, I want to thanks everyone who contributed to this crate. It's really awesome and it has clean so much code for me.

However, there's some features missing for me that I will implements after the holidays, such as an async support. And since all my code pass through clippy, denying all warnings, I began by that.

Happy holidays to you all !

PS: rustfmt have also been pass because it is configure on my IDE to run it on save, that makes more changes I have anticipated. Sorry for that.


Keats (approved):

Keats:
Thanks!

PR #181, opened 2022-01-06
Give possibility to pass `arg` to `schema` validator

amoriarty:
Hello everyone.

I add the possibility to pass `arg` to the `schema` validator, the same way it is done for the `custom` validator.

Once again, I'm not sure it was the way to do it, but at least it is there.


Keats:
I will need to look a this codebase again before doing a proper review, to be honest I don't remember much about it.

Keats (approved):

PR #186, opened 2022-02-02
feat: rocket feature for rocket json validation

somehowchris:
Lovely crate,

I would like to create a validation extension for rocket which has a type called `Json`. Since it's kind of a tricky one I would like to have type `T` validated for `Json<T>`.

The code is hidden behind the feature flag `rocket` and doesn't include default features of rocket, only the needed basics. I didn't meantion it in some documentations due to it being a "behind the scenes" thing as Request guards and much more would be needed to be implemented to work with rocket. As you can see [here](https://github.com/somehowchris/rocket-validation)

I'm opening this PR since rockets typing system doesn't allow for implementsions in third party crates


Keats:
I would prefer avoid having framework specific code in here. It seems to be an issue in Rocket that you cannot implement things like that separately :/

Issue #182, opened 2022-01-08
an attribute for skipping a struct field validation


coderedart:
I would like to skip a struct field which is just a `[f32; 3]` and there's no validation to be performed. I was wondering if it is possible to have a attribute like `#[serde(skip)]` which can skip validation for a particular struct field.

eg: `#[validate(skip)]`

Keats:
Validation is only done if you are setting an attribute on the field so there would be nothing to skip?

coderedart:
oh, but it gives me an error when i try to derive validation on a struct with `[f32; 3]` field.
```rust
use validator::Validate;

#[derive(Debug, Clone, Validate)]
pub struct Thing {
    pub pos: [f32; 3],
}```

just checked with a fresh crate just in case. I get the following error with `validator = {version = "*", features = ["derive"]}
` in cargo.toml.
```
➜  othing git:(master) ✗ cargo check
    Checking othing v0.1.0 (/home/user/programming/othing)
error: Type `[f32 ; 3]` of field `pos` not supported
 --> src/lib.rs:7:14
  |
7 |     pub pos: [f32; 3],
  |              ^^^^^^^^

error: could not compile `othing` due to previous error
```

**Comment by Keats** [comment]
Ah I see the issue. It should just ignore fields it doesn't recognize instead of erroring if there are no validators set on them.

**Comment by IniterWorker** [comment]
I'm having the same probleme here. Do you have a workaround or a solution? Maybe I can contribute.

```rust
error: Type `[u8 ; 11]` of field `reserved` not supported
  --> libcmd/src/lib.rs:88:19
   |
88 |     pub reserved: [u8; 11],
   |                   ^^^^^^^^
```

Keats:
No solutions yet, someone will need to go in the proc macro code, ctrl+F "not supported" and skip it instead of erroring if there are not validators set

PR #189, opened 2022-02-17
Allow unsupported_field_type to be ignored

IniterWorker:
@Keats,

First of all, It is my first time dealing with macro in rust.

Maybe it is not the best way to do it. But, I created a new skip validator.
- Alternative is to ignore at compile time (inside the macro).

I would like some code review on:
- General skip purpose
- Necessary/Required tests

If you have any hint, you are welcome ;-)

Related to: https://github.com/Keats/validator/issues/182

### DIFF
--- a/validator_derive/src/lib.rs
+++ b/validator_derive/src/lib.rs
@@ -342,12 +342,7 @@ fn find_fields_type(fields: &[syn::Field]) -> HashMap<String, String> {
             _ => {
                 let mut field_type = proc_macro2::TokenStream::new();
                 field.ty.to_tokens(&mut field_type);
-                abort!(
-                    field.ty.span(),
-                    "Type `{}` of field `{}` not supported",
-                    field_type,
-                    field_ident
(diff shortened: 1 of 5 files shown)


Keats:
It's not really a validator but more an attribute so I wouldn't go that way. I think it's fine to ignore all fields we cannot handle as long as they don't have a validator on them

IniterWorker:
Are you asking to remove the abort?

https://github.com/Keats/validator/blob/f49848545ee048a4b69be68463ca4eb40725067f/validator_derive/src/lib.rs#L316

Keats:
I think yes, according to https://github.com/Keats/validator/blob/f49848545ee048a4b69be68463ca4eb40725067f/validator_derive/src/lib.rs#L284-L285 it's only for the must_match so ideally we would only check fields with validators set

IniterWorker:
@Keats, maybe something like that can be a beginning?

PR #192, opened 2022-02-21
Add `sensitive` validator meta parameter

### LINKED ISSUE #157: errors returned by custom validators can't specify their own value parameter
I have the following custom validator:
```rust
/// A custom password validator that does not leak the password in the error it returns.
pub fn validate_password(password: &String) -> Result<(), ValidationError> {
    let is_valid = validator::validate_length(password, Some(10), Some(100), None);
    if is_valid {
        return Ok(());
    }

    let mut error = ValidationError::new("length");
    error.add_param("value".into(), &"");
    error.add_param("max".into(), &100);
    error.add_param("min".into(), &10);

    Err(error)
}
```

I would expect the error returned by `validate()` to contain an empty string as a value. However, the error contains the actual password.

I can provide a PR containing a test for reproduction if you like, however I'm not sure how to fix the problem itself :)

mdtusz:
Often, struct fields contain sensitive data which should never be displayed once the struct has been initially constructed or user input has been parsed (e.g. passwords and credit card numbers). This is especially problematic when logging validation errors as the invalid values may be potentially leaked in plaintext to logs.
    
This change adds a boolean `sensitive` validator "meta parameter" similar to the `message` and `code` parameters. This marks a field as sensitive such that the value should not be included in validation errors.

Resolves #157

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ match signup_data.validate() {
 
 A validation on an `Option<_>` field will be executed on the contained type if the option is `Some`. The `validate()`
  method returns a `Result<(), ValidationErrors>`. In the case of an invalid result, the `ValidationErrors` instance includes
-a map of errors keyed against the struct's field names. Errors may be represented in three ways, as described by the 
+a map of errors keyed against the struct's field names. Errors may be represented in three ways, as described by the
 `ValidationErrorsKind` enum:
 
 ```rust
@@ -339,13 +339,14 @@ if an error happened while validating the struct fields.
 
 Any error on the struct level validation will appear in the key `__all__` of the hashmap of errors.
 
-## Message and code
+## `message`, `code`, and `sensitive`
 
-Each validator can take 2 optional arguments in addition to their own arguments:
+Each validator can take 3 optional arguments in addition to their own arguments:
 
 - `message`: a message to go with the error, for example if you want to do i18n
 - `code`: each validator has a default error code (for example the `regex` validator code is `regex`) but it can be overriden
 if necessary, mainly needed for the `custom` validator
+- `sensitive`: when set to `true`, the field value will be omitted from validation errors, and can be used for fields with sensitive data such as passwords
 
 Note that these arguments can't be applied to nested validation calls with `#[validate]`.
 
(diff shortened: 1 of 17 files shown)


mdtusz:
On second thought, I think this may be better to have as a parameter equivalent to `required` which can just be in the top level of the macro usage. E.g.

```rs
#[validate(required, sensitive, length(min = 8))]
```

This will require some revisions, but should ultimately operate in a fairly similar way. Thoughts?

Keats:
+1 for having it at the top level

Keats:
Honestly it's been years since I touched the proc macro part but it will require some refactoring most likely yes

IniterWorker:
If we go with the current state of the art, this approach is good enough. 

But, if we want to get in bed with a **top-level approach**, we should consider two things that I think we should debate in another issue.
- Create a **shared state** in `StructValidation` at Struct's top level.
- Create a **shared state** in `FieldValidation` at Field's top level.

From **Validator's level**, we have access to **merged states**. 

**Pros:** 
- You will be able to access more information at the Validator's level.
- You should be able to require all fields from `StructValidation`,
- You should provide an easy way to implement and share new states/flags.

**Cons:**
- The validator shares a common set of features required to be implemented upfront (required, sensitive...).
- We increase the complexity.

mdtusz:
Apologies for abandoning this for such a long time. I'll rebase and see what I can do to get a top level parameter working. I think it's definitely the superior approach and much more ergonomic - there's essentially no use cases where you would consider a property sensitive for _some_ validation strategies, but not others. 

This may balloon into more structural changes though as @IniterWorker pointed out, so we'll want to open a few new issues around getting those changes made first, then making the required changes for `sensitive`, `required` etc.

Keats:
Honestly I'm thinking that between https://github.com/Keats/validator/issues/205 and this, it's probably easier to start from scratch. Seeing how many more features this crate now has compared to the PoC it was initially, it can probably be cleaned up quite a bit.

IniterWorker:
@Keats, I second #205 and the fact we must refactor the core  to add a `sensitive` validator or any new validator.

mdtusz:
Closing this as it is now incompatible with how the crate is implemented. I may be able to re-open with compatible changes in the next few weeks, but nobody hold their breath!

PR #203, opened 2022-03-24
Better support for maps

krojew:


Keats:
What do you mean by better support? If there are new features there should be some tests to go with it

krojew:
By better support I mean more general support for any map with any hasher, along with nested validation. You're right about tests - I'll add some.

Keats (approved):

PR #208, opened 2022-04-29
Remove transitive proc-macro dep from validator crate.

rosslannen:
Alternative solution to https://github.com/Keats/validator/pull/206. Instead of completely removing dependency on crate, just remove transitive dependency on `proc_macro` crate.


rosslannen:
This has yet to be fully tested, so don't do anything until that is complete.

PR #210, opened 2022-05-08
added message for required

pintariching:
When passing a message or code to the `required` attribute I get the following errors:
`unexpected list validator: "required"` and `no method named `validate` found for struct `can_specify_message_for_required::TestStruct` in the current scope`

I guess the `required` attribute doesn't take in any additional arguments? I'm still figuring out how to add that, here I just copied the tests from other attributes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
         include:
           - build: pinned
             os: ubuntu-20.04
-            rust: 1.46.0
+            rust: 1.47.0
           - build: stable
             os: ubuntu-20.04
             rust: stable
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
         include:
           - build: pinned
             os: ubuntu-20.04
-            rust: 1.51.0
+            rust: 1.56.0
           - build: stable
             os: ubuntu-20.04
             rust: stable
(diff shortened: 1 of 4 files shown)


Keats:
Looks that way indeed.

pintariching:
Okay I think I managed to do it. It seems like the "required" validator was missing from two match statements.

pintariching:
All the tests pass and you can now pass a message and a code to the required attribute. Should I close this pull request?

Keats:
My bad, it's fine. Looks like we just need to change the pinned Rust version.

pintariching:
To which version should the pinned version be changed?

Keats:
To the first version that passes the CI, you can usually infer it by seeing the error and googling around.

pintariching:
Okay it seems that the crate `hashbrown v0.12.1` is using the 2021 edition of rust, which is stabilized in rust 1.56.1. Should this be the pinned version?

Keats:
Looks that way! Not sure how exactly we're depending on it, I'll have to double check.

pintariching:
`indexmap v1.9.1` depends on `hashbrown v0.12.1`. Here's their [RELEASES.md](https://github.com/bluss/indexmap/blob/master/RELEASES.md) where they specify that the rust version should be at least 1.56.1 if you don't want to downgrade to `hashbrown v0.12.0`

Keats (approved):

PR #211, opened 2022-05-09
Email length validation before regex

### LINKED ISSUE #209: Check length in email-validation
There seems to be a maximum to the allowed number of characters in an email.

According to my current understanding, doing regex-checks on very long email-addresses might be expensive. Checking the length before going to regex would be a faster way of detecting malicious email-strings & preventing DOS attacks.

In any case, doing an additional length-check would improve recognition of impossible emails.

A few links that might be helpful:
- https://stackoverflow.com/questions/386294/what-is-the-maximum-length-of-a-valid-email-address#574698
- https://www.lifewire.com/is-email-address-length-limited-1171110
- https://blog.moonmail.io/what-is-the-maximum-length-of-a-valid-email-address-f712c6c4bc93

pintariching:
I added validation for emails before doing regex validation on the input. However these RFC standards are a bit confusing and I'm unsure what lengths to set each part of the address to.

The regex, I think, validates the lengths like [??].[??]@[63].[63] 

Is the length of the user part being validated?

Fixes #209

### DIFF
--- a/validator/src/validation/email.rs
+++ b/validator/src/validation/email.rs
@@ -33,13 +33,13 @@ where
     let user_part = parts[1];
     let domain_part = parts[0];
 
-	// validate the length of each part of the email, BEFORE doing the regex
-	// according to RFC5321 the max length of the local part is 64 characters
-	// and the max length of the domain part is 255 characters
-	// https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.3.1.1
-	if user_part.length() > 64 || domain_part.length() > 255 {
-		return false;
-	}
+    // validate the length of each part of the email, BEFORE doing the regex
+    // according to RFC5321 the max length of the local part is 64 characters
+    // and the max length of the domain part is 255 characters
+    // https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.3.1.1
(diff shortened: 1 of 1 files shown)


pintariching:
Or should the length of each part be an argument that can be optionaly passed in when creating the validation like so:
`#[validate(email(user_part_length = 63, domain_part_length = 255))]`

Glitchy-Tozier:
1. Thank you for creating this PR!
2. You could add "`closes #211`" to the initial comment to automatically close the issue when the PR gets accepted.

> Or should the length of each part be an argument that can be optionaly passed in when creating the validation like so:
`#[validate(email(user_part_length = 63, domain_part_length = 255))]`

I don't think this is needed. If emails are tested accordingly to an accepted standart, there shouldn't be a need for detailed customizability.

> However these RFC standards are a bit confusing and I'm unsure what lengths to set each part of the address to.

I also find it confusing. Some of the posts (see #211) suggest that the initial "320" was an error, and that it's actually "64 for username, 254 for full address (including the username!)

> So, when creating a new email address, limit your username to 64 or fewer characters, and the total number of characters to 254.

I haven't spent enough time reading official sites though, so I'm not sure which interpretation is correct.

PR #215, opened 2022-05-23
Added clippy suggestions

rakshith-ravi:
This PR is mostly just a code cleanup.

Clippy was complaining about a few things when I ran it, and they've been fixed in this PR. Most of them were `assert_eq!` being used with a `true` or `false` value, which has been replaced with `assert!`.

If we'd rather have the explicit true or false values, let me know and I'll revert those changes


Keats:
You went a bit too fast with the search&replace, some of those are assuming the false value

rakshith-ravi:
Yeah I'm not happy with that commit message. Stupid me. Let me rewrite that.

Keats:
Thanks!

PR #224, opened 2022-07-08
feat: add nested support for slices and most list/iterable types

Tuetuopay:
This now support nested validation for the following types:
- `[T; N]`
- `&[T]`
- `Option<[T; N]>`
- `&HashMap`, `&BTreeMap`, `&FxHashMap`, `&FnvHashMap`
- `HashSet` and `&HashSet`
- `BTreeSet` and `&BTreeSet`
- `IndexMap` and `&IndexMap`
- `IndexSet` and `&IndexSet`

Sadly, I was not able to get `Option<&T>` working properly because of lifetimes and how `_type` is generated inside of `FieldQuoter`: the space disappears between the lifetime and the inner type, making it impossible to extract the real type. (e.g. `Option<&'a HashSet<Child>>` becomes `Option<&'aHashSet<Child>>`, leading to the inner type `aHashSet`)

### DIFF
--- a/validator/src/traits.rs
+++ b/validator/src/traits.rs
@@ -50,6 +50,24 @@ impl<'a, T> HasLen for &'a Vec<T> {
     }
 }
 
+impl<T> HasLen for &[T] {
+    fn length(&self) -> u64 {
+        self.len() as u64
+    }
+}
+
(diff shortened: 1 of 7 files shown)


Keats (approved):

Keats:
Thanks!

Issue #205, opened 2022-04-21
Use traits for validation rules


tqwewe:
When trying to derive `Validate` on a struct, it seems like the derive macro filters what types you can use manually.

Wouldn't it make more sense to instead have a trait for each validation rule, allowing users to implement validation rules for their custom types?

For example, this does not work:
```rust
#[derive(Validate)]
struct Data {
  #[validate(length(min = 5, message = "must have 5+ characters"))]
  name: Mutable<String>,
}
```

But I'd like to be able to do something like this to resolve it:
```rust
impl validator::rules::ValidateLength for Mutable<String> {
  fn validate_length(&self, min: Option<u64>, max: Option<u64>) -> Result<(), ValidationError> {
    // ...
  }
}
```

Keats:
How would you the trait look like? I don't think it would work without variadic generics

tqwewe:
I'm not sure why generics may be needed.
I was imagining something like:
```rust
trait ValidateLength {
  fn validate_length(&self, min: Option<u64>, max: Option<u64>) -> bool;
}

trait ValidateEmail {
  fn validate_email(&self) -> bool;
}

trait ValidateUrl {
  fn validate_url(&self) -> bool;
}

trait ValidateContains {
  fn validate_contains(&self, s: &str) -> bool;
}

...
```
And these traits would be implemented for some standard types where appropriate, eg `impl ValidateEmail for String`, etc.

Keats:
Yeah my bad I was thinking the wrong thing yesterday. This would be nice but how do you envision it working in the proc macro? You still need to search for patterns in the fields types to know whether they are supported or try to di it anyway and get compilation errors in invisible generated code which isn't great.

pintariching:
I have opened #225 to see if I'm going in the right direction. For now I've just implemented if for length`

PR #225, opened 2022-07-09
Trait validation

pintariching:
I've implemented trait validation as mentioned in #205.

For now I've just implemented it for the `length` validation just to see if I'm on the right track. I've also copied the tests just to be sure it works.

Now if you want length validation for a custom type you just have to write an `impl` block for it like
```
impl ValidateLength for CustomType {
    fn validate_length(self, min: Option<u64>, max: Option<u64>, equal: Option<u64>) -> bool {
        // write length validation in here 
    }
}
```


validator/src/validation/length.rs
```diff
@@ -32,11 +34,33 @@ pub fn validate_length<T: HasLen>(
     true
 }
 
+pub trait ValidateLenght: HasLen {
```
tqwewe, line 37: It seems like if the user wants to implement `ValidateLength` on their own type, they'd need to also implement `HaLen`?

Might it be simpler if `HasLen` is not involved?

Keats: Typo on the name and it shouldn't use `HasLen` at all. Then we can just create a macro to generate all the impls we want for each type. With this trait, HasLen doesn't exist anymore

Keats (changes requested):

pintariching:
Okay I changed the trait and updated the implementations. The length extraction has to be done inside the `impl` for every type out there. I just added the temporary function `validate_length_for_trait` so the crate still builds.

pintariching:
I tried making a macro with `macro_rules!` for quickly implementing for all the types but my macro-fu is not on that level. I'm not sure how to specify lifetimes so the macro makes something like this
```
impl<'a> HasLen for &'a String {
    fn length(&self) -> u64 {
        self.chars().count() as u64
    }
}
```
from something like this 
```
impl_validate_length!(&'a String)
```

pintariching:
I have implemented all the impls that had the `HasLen` trait by hand. I'm unsure how to make a macro. I also added a derive test `can_validate_custom_impl_for_length` that shows the functionality that I managed to get to work. Right now you have to implement the length validation by yourself unless `validator::length::validate_length_for_trait` is made public

validator/src/validation/length.rs
```diff
@@ -32,11 +42,126 @@ pub fn validate_length<T: HasLen>(
     true
 }
 
+pub trait ValidateLength {
+    fn validate_length(&self, min: Option<u64>, max: Option<u64>, equal: Option<u64>) -> bool;
+}
+
+impl ValidateLength for String {
+    fn validate_length(&self, min: Option<u64>, max: Option<u64>, equal: Option<u64>) -> bool {
+        validate_length_for_trait(self.chars().count() as u64, min, max, equal)
```
tqwewe, line 51: What made you decide to go with `self.chars().count()` instead of `self.len()`?
And why use u64 everywhere instead of `usize`?

pintariching: I used `self.chars().count()` because the `HasLen` trait used them too and I didn't really think about it. The same is for `u64` instead of `usize`. The `validate_length` function used `Option<u64>`. I can refactor the trait to use `self.len()` and `usize` if it makes any difference.

pintariching:
Now I'm unsure what path to go. Do we let users implement the validation logic by themselves like so 
```
#[derive(Debug, Serialize)]
struct CustomString(String);

impl validator::ValidateLength for &CustomString {
    fn validate_length(&self, min: Option<u64>, max: Option<u64>, equal: Option<u64>) -> bool {
        let length = self.0.chars().count() as u64;

        if let Some(eq) = equal {
            return length == eq;
        } else {
            if let Some(m) = min {
	    if length < m {
	        return false;
	    }
        }
        if let Some(m) = max {
	    if length > m {
                return false;
	        }
            }
        }
        true
    }
}
```
This fees like cumebrsome and I would prefer if the user just has to implement a `length` function like 
```
#[derive(Debug, Serialize)]
struct CustomString(String);

impl validator::ValidateLength for &CustomString {
    fn length(&self) -> u64 {
        self.0.chars().count() as u64
    }
}
```

And maybe leave the first example for the `custom` validator?

validator/src/validation/length.rs
Keats: Strings should use `self.chars().count()`

Keats:
I don't see what this `length` method gives us though? It's not validating anything

validator/src/validation/length.rs
tqwewe: @Keats interesting, what is the advantage of that over `.len()`?

Keats:
```rs
fn main() {
    let txt = "日本語";
    println!("len {:?} vs chars {:?}", txt.len(), txt.chars().count());
}
```

will print

```
len 9 vs chars 3
```

In this case no one reasonable will think of len as number of bytes used (`.len()`) vs number of characters, which is what people think of when they think of a len for a string.

tqwewe: Wow very interesting! Makes sense to use chars count then

Keats:
I think that looks good

validator/src/validation/email.rs
```diff
@@ -73,6 +42,68 @@ fn validate_domain_part(domain_part: &str) -> bool {
...
+            return match domain_to_ascii(domain_part) {
+                Ok(d) => validate_domain_part(&d),
+                Err(_) => false,
+            };
+        }
+
+        true
+    }
+
+    fn to_email_string(&self) -> String;
```
Keats, line 80: That implies allocation, it should return a `&str` instead

validator/src/validation/length.rs
```diff
@@ -1,42 +1,164 @@
-use crate::traits::HasLen;
+use std::{borrow::Cow, collections::{HashMap, HashSet, BTreeMap, BTreeSet}};
+
+#[cfg(feature = "indexmap")]
+use indexmap::{IndexMap, IndexSet};
```
Keats, line 4: Thoughts on letting crates handle validator support themselves like they do for serde?

validator/src/lib.rs
```diff
@@ -73,17 +73,17 @@ mod validation;
...
-pub use validation::email::validate_email;
+pub use validation::email::{validate_email, ValidateEmail};
 pub use validation::ip::{validate_ip, validate_ip_v4, validate_ip_v6};
-pub use validation::length::validate_length;
+pub use validation::length::{validate_length, ValidateLength};
 pub use validation::must_match::validate_must_match;
 #[cfg(feature = "unic")]
 pub use validation::non_control_character::validate_non_control_character;
 #[cfg(feature = "phone")]
 pub use validation::phone::validate_phone;
```
Keats, line 83: I'm tempted to remove that one from validator, it's currently too generic to be easily usable and can easily be added manually if needed.

Keats (changes requested):

validator/src/validation/email.rs
pintariching: Returning a `&str` is impossible if I want to return a formatted string for example 
```
fn to_email_string(&self) -> &str {
    format!("{}@{}", self.user_part, self.domain_part).as_str()
}
```
How about returning Cow<str>?

validator/src/lib.rs
pintariching: Maybe leaving an example on how to add it with a custom validator after removing it would be a good idea?

validator/src/validation/email.rs
Keats: Cow is fine then

svenstaro:
What's the status on this? Is there anything that can be picked up here?

Keats:
Yes that would be great. I started to have a look at the derive crate and it seems nicer to use than 7 years ago. It also made me realise we were not handling the rename_all of serde afaik so that rename feature kinda sucked but no one complained? Maybe it's not super useful in the end.

The library itself should be that PR but I don't know whether we should do it in 2 phases (validation as trait -> derive rewrite later) or try to do it in one go.
I would probably go with the 2 phases since it would be too much work to do it in a single phase.

I've created a `next` branch so we can work on it without causing issues on master

PR #226, opened 2022-07-11
Add schemars integration

DXist:
This PR adds integration with [schemars](https://github.com/GREsau/schemars)

The integration derives `JsonSchema` trait for `ValidationErrors`, `ValidationErrorKind` and `ValidationError` types.

This allows to include `ValidationErrors` in OpenAPI operation definitions as result types. Example integration code is included as an integration test that uses `rocket` and `rocket_okapi` to define an API method that takes validated input result and returns `OperationError::ValidationErrors` error variant in case of validation failure.

Without this integration I used newtype wrappers for the mentioned types + From definitions that did several field clones. With integration no extra memory allocations is required.

Below is a visualisation of an API error type that has recursive `ValidationErrors` variant. The visualisation is done by [rapidoc](https://rapidocweb.com/).
Снимок экрана 2022-07-11 в 21 06 31

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
         include:
           - build: pinned
             os: ubuntu-20.04
-            rust: 1.56.1
+            rust: 1.57.0
           - build: stable
             os: ubuntu-20.04
             rust: stable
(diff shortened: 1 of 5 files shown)


DXist:
Unfortunately Rust 1.57 is required to run test for this feature.

IniterWorker:
Hi,

I had fast reviewed @DXist's commits in #226. Impl `schemars` is good to me. But what are your thoughts @Keats and @krojew, about tests/examples with a hard coupling on `rocket` and `rocket_okapi`. 
Perhaps too hard coupled on third-party libs `rocket` and `rocket_okapi`?

DXist:
Hi,

Thank you for feedback!

I think it's possible to introduce a separate crate in this repository to run integration tests. And isolate test dependencies there.

PR #227, opened 2022-07-21
Accept const variable in message

### LINKED ISSUE #142: Is it possible to make the the `message` be a variable?
Is it possible to make the the `message` be a variable?

For example I have this:

```
#[derive(Deserialize, Serialize, Validate, Debug)]
struct Auth {
    #[validate(length(min = 3, max = 21, message = "Username must be 3-21 characters long alphanumeric"), regex(path = "ALPHANUMERIC_REGEX",
    message = "Username must be 3-21 characters long alphanumeric"))]
    username: String,
    #[validate(length(min = 6, max = 100, message = "Password must be 6-100 characters long"))]
    password: String,
    #[validate(custom = "validate_authtype")]
    authtype: String,
}
```

Notice how I am having to repeat the string `"Username must be 3-21 characters long alphanumeric"` twice - once for the length and once for alphanumeric regex. Is it possible to put a `const` variable instead to avoid the repetition? Or some custom function for the message - this can be helpful for localization purposes too.

pintariching:
I've added an option to accept variables as a message. To pass a variable you have to prepend it with `"CONST:"`. It looks like this

```
const TOO_LONG_MESSAGE: &str = "value is too big";

#[derive(Debug, Validate)]
struct TestStruct {
    #[validate(range(max = 10, message = "CONST:TOO_LONG_MESSAGE"))]
    val: usize,
}
```
Closes #142

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Changelog
 
+## 0.16.0 (2022/06/27)
+
+- Allow passing code/message to `required`
+- Add `does_not_contain` validator
+- Check email length before validating it
+
 ## 0.15.0 (2022/05/03)
 
 - Allow passing args to schema validator
--- a/README.md
+++ b/README.md
@@ -218,8 +218,8 @@ Examples:
 ```
 
 ### does_not_contain
-Pretty much the opposite of contains, provided just for ease-of-use. Tests whether the string does not contain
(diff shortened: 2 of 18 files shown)


Keats:
I think it should be a different field instead of overloading with some magic string. Let's hold on to that though, we might have to rewrite the proc macro to make it easier to handle things like sensitive parameters.

PR #232, opened 2022-08-28
Add additional badges to README.md

wolthom:
This change adds additional badges to the README.md file for quicker navigation between crates.io / github / docs.rs.

The structure is shamelessly copied from SQLx (that was the first repo I thought of to check).


Keats:
Thanks!

PR #235, opened 2022-10-12
Improve Display Impl - Each ValidatorError should be on its own line

aarashy:
The current impl for Display doesn't break newlines frequently enough.

The display output of the list variant of `ValidationErrorsKind` currently looks like the following:
```
my_field[0].subfield: Validation error: Some error. [{"value": String("some-value")}]my_field[1].subfield: Validation error: Some error. [{"value": String("some-value-2")}]
other_field[0].other: Validation error: Other error [{"value": String("other-value")}]
```
It only breaks line when the top-level field of the `ValidationErrors` changes. I would prefer for it to look like the following:

```
my_field[0].subfield: Validation error: Some error. [{"value": String("some-value")}]
my_field[1].subfield: Validation error: Some error. [{"value": String("some-value-2")}]

other_field[0].other: Validation error: Other error [{"value": String("other-value")}]
```
It seems pretty clear that separate field-level errors belong on their own line - having multiple on the same line always looks bad. Having multiple line-breaks when the top-level field changes doesn't feel like a regression either, but we could remove the existing `writeln!` call if we wanted a maximum of 1 new line after every error display.

### DIFF
--- a/validator/src/display_impl.rs
+++ b/validator/src/display_impl.rs
@@ -5,9 +5,9 @@ use crate::{ValidationError, ValidationErrors, ValidationErrorsKind};
 impl fmt::Display for ValidationError {
     fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
         if let Some(msg) = self.message.as_ref() {
-            write!(fmt, "{}", msg)
+            writeln!(fmt, "{}", msg)
         } else {
-            write!(fmt, "Validation error: {} [{:?}]", self.code, self.params)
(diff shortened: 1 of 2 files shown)


Keats (approved):

aarashy:
Hi, thanks for the approval - the issue with CI seems unrelated to this PR, right?

Keats:
It seems unrelated yes

aarashy:
Would you like me to re-run the CI or will you merge this later?

Keats:
The concept probably, but ideally there will be an error rewrite for the next version to make it simpler.

PR #237, opened 2022-12-28
Prevent possible collision with struct's own validate() method

KonishchevDmitry:


validator_derive/src/quoting.rs
```diff
@@ -491,7 +491,8 @@ pub fn quote_regex_validation(
 pub fn quote_nested_validation(field_quoter: &FieldQuoter) -> proc_macro2::TokenStream {
     let field_name = &field_quoter.name;
     let validator_field = field_quoter.quote_validator_field();
-    let quoted = quote!(result = ::validator::ValidationErrors::merge(result, #field_name, #validator_field.validate()););
+    let quoted = quote!(result = ::validator::ValidationErrors::merge(
+        result, #field_name, ::validator::Validate::validate(&#validator_field)););
```
kyrias, line 495: This line wrapping is a bit strange.  How about something like this to make it more readable?

```suggestion
    let quoted = quote!(
        result = ::validator::ValidationErrors::merge(
            result, #field_name, ::validator::Validate::validate(&#validator_field)
        );
    );
```

Keats: that's a nice change, but otherwise i'll just keep whatever `cargo fmt` does after merging

kyrias: For the record, `cargo fmt` won't do anything with it since it's a macro where it doesn't know what the expected format is.

Keats (approved):

Keats:
Still an issue with nested validators but the derive has been rewritten. The same change does not seem to compile on the newer code.

PR #245, opened 2023-03-20
add length_utf16 validator

DXist:
This PR adds length_utf16 validator.

My project exposes data from Salesforce via JsonSchema based API. I want to validate field lengths in the same way as Salesforce does - by counting UTF16 characters.

UTF16 is used for Unicode string representation in JavaScript, Java and Salesforce APEX. I think this validator could be useful to others as well.

Should I wrap the implementation in optional feature `length_utf16` ?

### DIFF
--- a/validator_derive/src/quoting.rs
+++ b/validator_derive/src/quoting.rs
@@ -223,6 +223,66 @@ pub fn quote_length_validation(
     unreachable!()
 }
 
+pub fn quote_length_utf16_validation(
+    field_quoter: &FieldQuoter,
+    validation: &FieldValidation,
+) -> proc_macro2::TokenStream {
+    let field_name = &field_quoter.name;
+    let validator_param = field_quoter.quote_validator_param();
+
+    if let Validator::LengthUTF16 { min, max, equal } = &validation.validator {
+        let min_err_param_quoted = if let Some(v) = min {
+            let v = value_or_path_to_tokens(v);
+            quote!(err.add_param(::std::borrow::Cow::from("min"), &#v);)
+        } else {
+            quote!()
+        };
+        let max_err_param_quoted = if let Some(v) = max {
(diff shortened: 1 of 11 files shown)


Keats:
I don't think it makes sense to add that to the library, it's better added as a custom validator.

validator_derive/src/quoting.rs
```diff
@@ -223,6 +223,66 @@ pub fn quote_length_validation(
...
 }
 
+pub fn quote_length_utf16_validation(
+    field_quoter: &FieldQuoter,
+    validation: &FieldValidation,
+) -> proc_macro2::TokenStream {
+    let field_name = &field_quoter.name;
+    let validator_param = field_quoter.quote_validator_param();
+
+    if let Validator::LengthUTF16 { min, max, equal } = &validation.validator {
```
LeoniePhiline, line 233: For consistency with Rust code style, you might want to use `Utf` in identifiers.

E.g. https://doc.rust-lang.org/std/str/struct.EncodeUtf16.html

validator_types/src/lib.rs
```diff
@@ -81,6 +87,7 @@ impl Validator {
             Validator::Regex(_) => "regex",
             Validator::Range { .. } => "range",
             Validator::Length { .. } => "length",
+            Validator::LengthUTF16 { .. } => "length_utf16",
```
LeoniePhiline, line 90:
```suggestion
            Validator::LengthUtf16 { .. } => "length_utf16",
```

For consistency with Rust code style, you might want to use `Utf` in identifiers.

E.g. https://doc.rust-lang.org/std/str/struct.EncodeUtf16.html

DXist: I'll address code style suggestions if the approach with an extra builtin validator type is desired for the crate users.

We could gather more comments/thumbs up in the MR description for more feedback.

PR #253, opened 2023-05-26
Fully remove code for old phone validation

kyrias:
The phone validation code was partially removed in 897811a but parts were left in leading to tests not passing.

This also updates the expected compiler errors for two compile-fail cases.  There is one that is still failing, but I haven't had the time to look into it yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,8 +45,6 @@ use validator::{Validate, ValidationError};
 struct SignupData {
     #[validate(email)]
     mail: String,
-    #[validate(phone)]
-    phone: String,
     #[validate(url)]
     site: String,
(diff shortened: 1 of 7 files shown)


Keats (approved):
Thanks!

Keats:
Need to implement validate_email for Vec<u8> it seems for the last test

kyrias:
Oh right, it's to allow them to be implemented for custom types.  It definitely shouldn't be implemented for `Vec<u8>` but I guess this means that we'll just have to drop this compile fail test since if we allow any type to implement it, we _cannot_ have the derive macro give a nice error for this.

PR #254, opened 2023-05-26
Delete old not_a_string_type test

kyrias:
This test was checking that the derive macro assertion about the type not being one of the known string types was working, but now since we allow implementations of the email validation for third-party types there is no way for us to perform this check in the derive macro.

The same assertion is still used for the credit card and non-control character validators, but those will also soon be switched to be trait-based.


Keats:
I think we should probably remove all those checks no?

kyrias:
They should be removed when the same has been done for those two constraint types, but I feel like it's reasonable to have the test for them until that's done, though looking at them that should be quickly done, hm.

PR #257, opened 2023-05-29
Switch non-control-character validator to a trait

kyrias:


validator/src/lib.rs
```diff
@@ -75,7 +75,9 @@ pub use validation::ip::{validate_ip, validate_ip_v4, validate_ip_v6};
 pub use validation::length::{validate_length, ValidateLength};
 pub use validation::must_match::validate_must_match;
 #[cfg(feature = "unic")]
-pub use validation::non_control_character::validate_non_control_character;
+pub use validation::non_control_character::{
```
Keats, line 78: We could potentially make the trait not be gated behind the feature but then the default impl would be invalid... Not sure it's worth ti

Keats (changes requested):
Some legit CI errors it seems

kyrias:
> Some legit CI errors it seems

No, it's just that removing the assert function could only be done with the previous two PRs merged, since GitHub doesn't allow for stacking PRs across repos.

validator/src/lib.rs
kyrias: Hmm, why?

kyrias: I guess I could see that, though it's a bit strange to have a default function implementation be conditionally compiled, and would also need the blanket implementation to be conditionally compiled.

Keats: I agree. Let's not bother with it

PR #259, opened 2023-07-05
move `lazy_static` to `once_cell`

attila-lin:


Keats:
Thanks! Can you do the PR on this branch: https://github.com/Keats/validator/pull/249 ?

pintariching:
Have you looked at the `OnceCell` type in the std library? The functionality is the same as in the `once_cell` crate I think?

Keats:
> Have you looked at the OnceCell type in the std library? The functionality is the same as in the once_cell crate I think?

It is, but brand new. It depends which version of Rust we want to target.

pintariching:
The MSRV of `once_cell` is `1.56.0` and `OnceCell` was introduced in the std library in `1.70.0`. Either the MSRV is updated to `1.70.0` if the std library is used or to `1.56.0` if `once_cell` is used.

Or maybe have everything that uses the std `OnceCell` behind a feature and use `once_cell` by default if that makes sense?

Keats:
Since we're not planning to release straigh away, you can probably depend on 1.70 and use the one from std

Keats:
Done in the next branch. Not using the one from std since LazyLock is not stable yet

PR #260, opened 2023-07-13
Add starts_with validator

evaporei:


evaporei:
I just saw that the last PR had to point to the `next` branch, should I do the same?

Keats:
I think it would be better to hold on adding more validator until the rewrite is done

PR #261, opened 2023-07-23
Disable unnecessary/unused regex features

jirutka:
### Avoid unnecessary case-insensitive regexes

The case-insensitive flag `(?i)` requires regex with the unicode-case
feature enabled which significantly increases the binary size and makes
regex slower.

### Disable unnecessary/unused regex features

regex crate with default features enabled is huge, it significantly
increases the binary size and compile time.

Cargo features are additive, so once some feature is enabled anywhere in
the dependency graph, there's no way to disable it, i.e. one
(transitive) dependency infects the whole dependency graph.

Since `#[validate(regex = "...")]` takes a path to a static Regex
instance, not regex as a string, users have to add the `regex`
dependency to their Cargo.toml to use it. Therefore, this change
shouldn't break backward compatibility.

### DIFF
--- a/validator/Cargo.toml
+++ b/validator/Cargo.toml
@@ -12,7 +12,7 @@ readme = "../README.md"
 
 [dependencies]
 url = "2"
-regex = "1"
+regex = { version = "1", default-features = false, features = ["std"] }
 lazy_static = "1"
 idna = "0.2"
 serde = "1"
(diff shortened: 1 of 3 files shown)


jirutka:
> Can you make the PR on the next branch?

Ok, done.

Keats (approved):

jirutka:
Thanks for merging! When can I expect a new release?

Keats:
Not anytime soon, the plan is to rewrite the derive macro for the next version

jirutka:
If it takes longer, why don’t you just release the new version now and rewrite the macros in the next one?

Issue #241, opened 2023-02-13
Implement the same behavior as Option


affanshahid:
I have a type which behaves the same as an `Option` in that it wraps another type which might or might not be there. It has some extra logic that is specific to my app. I want to give it the same behavior that an `Option` has i.e only validate if the value is there and pass on all validation settings to the wrapped type. How should I go about achieving this.

biot023:
+1

biot023:
(In my case it's the `MaybeUndefined<T>` from async_graphql `InputObject`.)

Issue #236, opened 2022-12-24
Implement Validate for `Option<T: Validate>`


tsar-boomba:
Currently, option validation is done through the proc macro by checking if the type starts with `Option<` this is causing problems for me because I have to use the full path to option (`std::option::Option`) due to a bug in another crate. I may be missing something, but why not just do something like this:

```rs
impl<T: Validate> for Option<T> {
    pub fn validate(&self) -> Result<(), ValidateError> {
        if let Some(inner) = self {
            inner.validate()
        }
    }
}
```

Other than this we could check if someone is using the full path when doing option unwrapping in the proc macro. I'd be willing to do a pr for either, just checking to see what aligns with the goals of this crate the most.

Keats:
Yes, we should move away from checking naively in the macro and let the compilation just fail imo.

gzp-crey:
Any update on this ?

Issue #191, opened 2022-02-21
support async validator


LyunKi:
In some cases, we may need an async validator .

IniterWorker:
Are you requesting to allow `async fn` call througth the  `custom` field in `validate` macro ?

bradleyess:
@IniterWorker - I'm here to suggest that async fn's can be used in `custom` derive annotations. I'm not sure where to start but would be happy to collaborate on it.

paramako:
any progress on this?

Keats:
None. A rewrite of the macro has landed on the next branch if someones wants to have a look.

pintariching:
We should also think about if we want to be async framework agnistic or support just `tokio` or `smol` or others or all of them with features.

DenuxPlays:
Any progress on this?
Or any workaround that works with this crate: https://github.com/rambler-digital-solutions/actix-web-validator ?
Supporting only non async methods makes this crate unusable for something like username validation if an async connection is used to the database.

digitalillusion:
This is a way to have async context inside the non async validator

```
use tokio::runtime::Handle;

pub struct ValidateContext {
    pub db: Database,
    pub handle: Handle 
}

#[derive(Debug, Validate)]
#[validate(schema(function = "CreateUserPayload::validate_with_context", use_context), context = "ValidateContext")]
pub struct CreateUserPayload {
    pub username: String,
}

impl CreateUserPayload {
    pub fn validate_with_context(payload: &CreateFlowStepPayload, context: &ValidateContext) -> Result<(), ValidationError> {
        let ValidateContext { prisma, handle } = context;
        if handle.block_on(UsersRepository::find_by_username(db, &payload.username)).is_some()) {
            Err(ValidationError::new("username_exists"))
        } else {
            Ok(())
        }
    }
}
```

Issue #172, opened 2021-10-04
Required doesn't accept message argument


simanacci:
The <i>readme</i> states that all validators accept the message argument but `required` does not.

> Each validator can take 2 optional arguments in addition to their own arguments:

```rust
#[derive(Debug, Deserialize, Validate)]
struct User {
    #[validate(required(message ="My Message"))]
    name: Option<String>,
}
```
```
error: unexpected list validator: "required"
  --> src/main.rs:38:16
   |
38 |     #[validate(required(message = "My Message"))]
   |                ^^^^^^^^
error: aborting due to previous error
```

pintariching:
this is fixed in #210 and is going to be fixed in 0.16.0 #219

PR #262, opened 2023-07-26
Rewrite derive macro implementation

pintariching:
A proposal on rewriting the derive macro implementation using the `darling` crate. It massively simplifies the macro. The only two validators I got up an working are the `length` and `email` just for show. Also there is no error handling for now.

Also I had to upgrade `syn` version to `2` as `darling` would work with version `1`.


Keats (commented):
Veeery nice

validator_derive/src/lib.rs
```diff
@@ -1,629 +1,164 @@
...
+// Structs to hold the validation information and to provide attributes
+// The name of a field here corresponds to an attribute like
+// #[validate(length(min = 1, max = 10, equal = 8, message = ""))]
+//                   ^^^      ^^^       ^^^^^      ^^^^^^^
+#[derive(Debug, Clone, FromMeta)]
+struct Length {
+    min: Option<u64>,
+    max: Option<u64>,
+    equal: Option<u64>,
+    message: Option<String>,
```
Keats, line 19: It's missing` code` as well. It is much nicer and waaaay less code, very nice.

Random thought: is it possible to parse all the attributes into a hashmap rather than a struct? I'm wondering if there would be a way for a user to register functions with validator and use them on the derive as if they were built-in. Maybe it's not worth it though.

pintariching: There is a `map` attribute for the macro that let's you specify a function to run after parsing the attributes like

```rust
#[derive(FromMeta)]
#[darling(map = "to_hashmap")]
struct Custom {
    par_1: Option<Expr>,
    par_2: Option<Expr>,
    par_3: Option<Expr>,
    #[darling(skip)]
    hashmap: HashMap<String, Expr>
}

fn to_hashmap(custom: Custom) -> HashMap<String, Expr> {}
```
I have to try it out, I'm not sure if this will work

Keats: It's only worth it if we have some kind of register of fns, and I'm not sure how to configure where to look that up... Maybe let's not worry about it for now?

pintariching: Very well. I'll implement the rest of the macro and get it up and running and possibly passing the tests and then worry about everything else.

validator/src/validation/length.rs
```diff
@@ -12,7 +12,7 @@ use indexmap::{IndexMap, IndexSet};
 /// If you apply it on String, don't forget that the length can be different
 /// from the number of visual characters for Unicode
 #[must_use]
-pub fn validate_length<T: ValidateLength>(
+pub fn validate_length<T: ValidateLength<u64>>(
```
Keats, line 15: Why?

pintariching: The function is kind of useless now that there's the trait I think. Also casting the `min`, `max` and `equal` values to `u64` might lead to issues when you for example cast a negative `i32` to `u64`. So I thought to add a generic, so that the `length` and all the parameters have to be the same type to compare them without runtime errors or weird casting issues.

Keats: We can remove it then I think?

pintariching: None of the functions will be used in the macro from the validator crate so they can be removed unless they're used anywhere else?

Keats: Yep they can be removed now that we have traits

pintariching:
I've rewritten the custom validation a bit. If you want arguments for your function you can specify them like:
```rust
#[derive(Validate)]
struct TestStruct {
    #[validate(custom(
        function = "valid_reference_with_lifetime",
        arg = "len: i64",
        arg = "something: bool",
        arg = "name: String"
    ))]
    value: String,
}

let test = TestStruct { name: "asdfg".to_string() };
let res = test.validate(TestStructArgs { len: 123, something: false, name: "Something".into() });
```
Also the `TestStructArgs` could be given a builder pattern if we wanna be fancy, so you get
```rust
TestStructArgs::build().len(123).something(false).name("Something").build()
```

How does this look? I'm now wrangling with lifetimes so for now you can only pass values and not references. Should I continue with this or go back to tuples like it's now?

Keats:
What do you think of the approach from garde: https://github.com/jprochazk/garde/tree/main#custom-validation ?

pintariching:
I think I like `garde`s way more, it feels simpler and more "proper". I would only add the `context` attribute in the `custom` attribute, so you can have multiple contexts for different fields and pass them in the `validate` function with a tuple and lifetimes and generics should also work for example

```rust
#[derive(Validate)]
struct Test {
    #[validate(custom(function = "custom_password", context = "PasswordContext")]
    password: String,
    #[validate(custom(function = "custom_name", context = "&mut NameContext<u32>")]
    name: String
}

struct PasswordContext { ... }
struct NameContext<T> { ... }

fn custom_password(value: &str, contex: PasswordContext) -> Result<(), ValidationErrors> { ... }
fn custom_name(value: &str, context: &mut NameContext<u32>) -> Result<(), ValidationErrors> { ... }

let test = Test { password: ..., name: ... };
test.validate((PaswordContext { ... }, { NameContext<u32> { ... }));
```

pintariching:
Ok I tried to do something similar to `garde` but couldn't get the lifetimes to parse with darling, so I tried a different way with closures and I think it's a bit better. Here's an example straight from the tests for convenience:

```rust
fn valid_fn(_: String, _arg: i32) -> Result<(), ValidationError> {
    Ok(())
}

#[derive(Validate)]
struct TestStruct {
    #[validate(custom)]
    value: String,
}

let test_struct = TestStruct { value: "Something".to_string() };
assert!(test_struct.validate(|v| valid_fn(v, 123)).is_ok());
```

The value gets cloned into the closure, that can be changed to a reference maybe with an attribute or to use a reference by default

Keats:
The issue is that it doesn't work with multiple custom function does it?

pintariching:
Check the `validator_derive_tests::custom_args` tests. I have one with multiple custom functions on multiple fields. But I guess I haven't tried multiple custom functions on a single field?

Keats:
No need to handle multiple custom on a single field i think, but different function on different fields should work

pintariching:
The only thing I'm unsure of is, I'm defining the `ValidateArgs` [inside the macro ](https://github.com/pintariching/validator/blob/427d386b86b672f565a92ea563c1aa41d05ad7bd/validator_derive/src/lib.rs#L327C12-L327C12), so it's not using the `validator::ValidateArgs` trait. With this, I think I could ditch the trait and just do an `impl` block for the struct with the `validate` function being in it.

pintariching:
Ok changed a lot of stuff, all the tests pass except `validator_derive_tests::complex::failed_validation_points_to_original_field_name`. `darling` doesn't seem to support this out of the box, so I might need to copy over the parsing from the previous version.

Keats:
Ignore that one for now, because we don't really support all the rename options from serde. Let's get everything working first and then see if we can point back to the original name later.

Keats:
Have you seen https://github.com/Keats/validator/issues/264 ?

pintariching:
This might require that we either support one or the other, or require that the regex is passed into the `validate` function as an argument like the `custom` validator.

In the macro, I don't think I can extract the type of a variable (`Regex` or `OnceLock<Regex>`) from the path `crate::REGEX`.

I added the `ValidateRegex` trait so another option would be to use generics instead of `Regex` like `T: Into<Regex>` but I'm not sure if `OnceLock` supports that trait

pintariching:
I have implemented a change to allow the use of either `OnceLock<Regex>`, `Mutex<OnceCell<Regex>>` or a `lazy_static` macro regex. The only change you have to do, is to dereference the type that is generated by `lazy_static`, but the others work fine. Also you have to specify the path to the static variable without double qoutes. Check [the tests for exapmles](https://github.com/pintariching/validator/blob/b6e1346e0690b961c5f01a8e483bb8f642101795/validator_derive_tests/tests/regex.rs#L17).

pintariching:
I still need to get the UI tests working and support renamed fields, however everything else seems to be working. Also just saw the checks and there seems to be a version mismatch for `hashbrown`.

Keats:
Thanks, I'll have a look asap! For the CI it might just be a bump needed for the pinned Rust version

pintariching:
Had to bump MSRV to 1.70.0, because we use `OnceCell` and `OnceLock`.

pintariching:
Yes it orders the closures in the tuple based on the order of fields. It puts them in a `Vec` and pushes any schema validation the struct has on the end.

Keats:
I think it's better to have a single context rather than multiple ones, otherwise nested custom is going to require the user to do their own book-keeping and easy to introduce mistakes when moving/adding/removing fields

Keats:
I think if you pass a single struct for all validation (which could be likely, eg a db connection and maybe some Option<String> for a specific field), you get nested custom for free?

pintariching:
I'm not entirely sure what you think. If we pass a single struct into the `validate()` function and have it used by all custom validations for the current struct and all the children? I see that [garde has a similar thing going on](https://docs.rs/garde/latest/garde/#custom-validation).

This could work for nested functions, but all custom functions would have to have it as an argument or accept multiple contexts I think.

Keats:
I was thinking of something like `garde` yes. Every custom function would then get it as argument

pintariching:
Ok I got something working. However now the call `validate(())` requires an empty tuple to be passed in everywhere, even if custom validation is not used. I'm not sure how to fix this.

Keats:
Can we do like currently and have both validate() and validate_with_args() and validate() would just call `validate_with_args(())`?

pintariching:
Ok I got it working. Now the only thing that's bugging me is, that I can't implement `ValidateNested` for both `T` and `&T`. Currently the test `nested::test_can_validate_option_fields_with_lifetime` doesn't pass because of this.

pintariching:
The custom validation with a given context is giving me quite a bit of trouble. It works now but is very finicky especially with generics and lifetimes. This all stems from the fact that we're implementing a trait, which can't accept different arguments. 
How about we ditch the `Validate` and `ValidateArgs` and impl directly on the struct. We can then name the custom arguments and you can much easier pass in everything you want with closures, ie:

```rust
#[derive(Validate)]
struct Test {
    #[validate(custom)]
    val: String
}

fn custom_validation(val: &String, context: &mut SomeContext, index: i32) -> Result<(), ValidationError> { ... }

/// This impl is autogenerated
impl Test {
    impl validate(&self, val_custom_closure: FnOnce(&String) -> Result<(), ValidationError> {
        val_custom_closure(self.val)
    }
}

let some_context = ...;
let index = ...;
// user gets to choose what arguments they want to use
test.validate(|val| custom_validation(val, some_context, index);
```

The argument fields are all named, so with multiple custom functions, you see the name of each parameter.

Also a builder pattern could be used something along the lines of
```rust
test
    .validation_builder()
    .val_custom_validation(|val| ...)
    .other_custom_validation(...)
    .validate();
```

Keats:
I don't like the builder pattern.
How about we do like currently and only allow a single lifetime?

pintariching:
Ok I think I got everything working except the `serde` original field name. The issues this should solve or partially solve are: #264, #241, #236, #205, possibly #191, #176, #172, #170, #83, maybe #73

Now I'm working on the errors and getting all the `compile-fail` and `run-pass` tests to work, but otherwise I think this could be used right now.

Keats:
I'll have a look asap

pintariching:
The UI tests now pass. I had to put them behind a feature to avoid running them on anything other than `stable` due to the errors that keep changing and reordering. Also for some reason the error produced by `compile-fail::range::wrong_type` test was different in the CI and locally, even though both use `stable`, so I removed it.

validator_derive_tests/tests/schema.rs
```diff
@@ -114,81 +117,103 @@ fn can_fail_schema_fn_validation() {
     assert_eq!(errs["__all__"][0].code, "meh");
 }
 
+// #[test]
+// fn can_fail_multiple_schema_fn_validation() {
```
Keats, line 121: can that test be enabled?

Keats (commented):
I admit i went a little fast on the review but i left some comments

validator/src/validation/regex.rs
```diff
@@ -0,0 +1,175 @@
+use std::borrow::Cow;
+use std::cell::OnceCell;
+use std::sync::{Arc, Mutex, OnceLock};
+
+use regex::Regex;
+
+pub trait IntoRegex {
+    fn into_regex(&self) -> Cow<Regex>;
```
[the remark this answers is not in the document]
pintariching: You're right, fixed it.

validator/src/validation/urls.rs
```diff
@@ -2,30 +2,143 @@ use std::borrow::Cow;
...
     fn validate_url(&self) -> bool {
-        Url::parse(&self.to_url_string()).is_ok()
+        if let Some(u) = self.to_url_string() {
+            Url::parse(&u).is_ok()
+        } else {
+            true
+        }
+    }
+
+    fn to_url_string(&self) -> Option<Cow<str>>;
```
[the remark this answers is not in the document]
pintariching: Fixed all traits to have this

validator/src/validation/urls.rs
```diff
@@ -2,30 +2,143 @@ use std::borrow::Cow;
...
+    fn to_url_string(&self) -> Option<Cow<str>> {
+        if let Some(u) = self {
+            if let Some(u) = u {
+                Some(u.clone())
+            } else {
+                None
+            }
+        } else {
+            None
+        }
```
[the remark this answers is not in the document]
pintariching: Clippy says it likes `self.as_ref().cloned().flatten()`

validator_derive/src/tokens/ip.rs
```diff
@@ -0,0 +1,45 @@
...
+        }
+    } else if let (None, Some(v6)) = (ip.v4, ip.v6) {
+        if v6 {
+            quote!(validate_ipv6())
+        } else {
+            quote!(validate_ip())
+        }
+    } else {
+        quote!(validate_ip())
+    };
```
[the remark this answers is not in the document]
pintariching: Fixed

validator_derive/src/types.rs
```diff
@@ -0,0 +1,214 @@
...
+    pub non_control_character: Option<Override<NonControlCharacter>>,
+    pub range: Option<Range>,
+    pub required: Option<Override<Required>>,
+    pub required_nested: Option<Override<Required>>,
+    pub url: Option<Override<Url>>,
+    pub regex: Option<Regex>,
+    pub custom: Option<Custom>,
+    pub skip: Option<bool>,
+    pub nested: Option<bool>,
+}
```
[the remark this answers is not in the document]
pintariching: I tried this at first but `darling` doesn't seem to support this out of the box as far as I've seen unless we write a custom parser I think.

validator_derive_tests/nested.rs
```diff
@@ -0,0 +1,2750 @@
+#![feature(prelude_import)]
```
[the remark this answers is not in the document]
pintariching: Looks like I forgot to remove the file when playing around with `cargo expand`. Fixed

validator_derive_tests/tests/compile-fail/length/unknown_arg.stderr
```diff
@@ -1,5 +1,5 @@
-error: Invalid attribute #[validate] on field `s`: unknown argument `eq` for validator `length` (it only has `min`, `max`, `equal`)
- --> $DIR/unknown_arg.rs:5:23
+error: Unknown field: `eq`. Did you mean `equal`?
```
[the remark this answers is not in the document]
pintariching: This is an error from `darling`. I think it defaults to field for everything?

validator_derive_tests/tests/complex.rs
```diff
@@ -23,21 +23,21 @@ fn validate_signup(data: &SignupData) -> Result<(), ValidationError> {
...
     site: String,
-    #[validate(length(min = 1), custom = "validate_unique_username")]
+    #[validate(length(min = 1), custom(function = validate_unique_username))]
     #[serde(rename = "firstName")]
     first_name: String,
     #[validate(range(min = 18, max = 20))]
     age: u32,
     #[validate]
-    card: Option<Card>,
+    _card: Option<Card>,
```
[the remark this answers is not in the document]
pintariching: Added `#[allow(unused)]` at the struct instead. Cargo was giving warnings because it's not used.

Keats:
Just a couple more tests that are currently commented out, can I merge and uncomment them or is there a reason they are commented out?

pintariching:
Sorry for the wait, forgot about this. No the tests don't pass currently if you uncomment them I think. I was thinking gating the `serde` original name behind a feature.

Keats:
ah it's all related to that feature? You can just remove it for now then

validator/src/types.rs
```diff
@@ -58,6 +58,22 @@ impl ValidationErrors {
         }
     }
 
+    pub fn merge_self(
```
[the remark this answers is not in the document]
pintariching: I had some trouble adding errors on nested validations so I whipped up this function. And I agree, the errors API should be redone.

pintariching:
Ok I removed the original field name from serde test and added back multiple schema validation.  I think that makes everything? Also the documentation should be updated, this PR is a breaking change after all.

Keats:
Let's update the docs later, we might have more things.

Thanks a lot!

PR #267, opened 2023-08-29
Add ValidateAll derive macro which automatically adds nested validation to all fields

davidlang42:
My suggestion here is to add a second derive macro `ValidateAll` which operates similarly to the original (and unmodified) `Validate` macro, except if any fields don't have any validation, they are assumed to have Nested validation (as if they had the attribute `#[validate]`.

The use case here is when you have a Struct with many fields, most of which are more nested Structs. Down a few layers you probably have a Struct with more typical validation, such as:

```
#[derive(Validate)]
struct Address {
    name: String,
    #[validate(range(max = 99))]
    house_number: u8
    street: String
    #[validate(phone)]
    phone_number: String
    suburb: String
}
```
but in the upper layers you have many fields, all of which need Nested validation, writing `#[validate]` on every 2nd line would be quite verbose. So instead of having this:
```
#[derive(Validate)]
struct Order {
    #[validate]
    customer_number: u16,
    #[validate]
    billing_address: Address,
    #[validate]
    postal_address: Address,
    #[validate(email)]
    email: String,
    backup_address: Address
    #[validate]
    item_type: Item,
    #[validate(range(min = 1))]
    item_quantity: u8,
    #[validate]
    payment_method: Payment,
    #[validate]
    referrer: MarketingSource
}
```
you can simply put:
```
#[derive(ValidateAll)]
struct Order {
    customer_number: CustomerId,
    billing_address: Address,
    postal_address: Address,
    #[validate(email)]
    email: String,
    backup_address: Address
    item_type: Item,
    #[validate(range(min = 1))]
    item_quantity: u8,
    payment_method: Payment,
    referrer: MarketingSource
}
```
However there is another, and much more important catch here. The first option is not just verbose but its also prone to missing a field, meaning you *think* you are validating the nested data but you actually aren't. If you look closely, you'll see I forgot the `#[validate]` attribute on `backup_address` in the first example, but its not obvious at all. Once I use ValidateAll instead, it not only does the nesting for me, but if any of the types don't implement Validate (for example `MarketingSource`) then I'll get a compiler error telling me I forgot something.

I can't help but think this will benefit many people in many scenarios, but also I'm not tied to the exact implementation. If you don't like the extra derive macro `ValidateAll` another idea I had was adding a struct level attribute like `#[validate(mandatory_on_all_fields)]` which makes the `Validate` derive macro do the same thing as above.

Also in this PR is a new field level attribute `#[validate(always_valid)]` for the odd time when you want to use `ValidateAll` but some fields genuinely don't need validation.

### DIFF
--- a/validator/src/lib.rs
+++ b/validator/src/lib.rs
@@ -91,3 +91,5 @@ pub use types::{ValidationError, ValidationErrors, ValidationErrorsKind};
 
 #[cfg(feature = "derive")]
 pub use validator_derive::Validate;
+#[cfg(feature = "derive")]
+pub use validator_derive::ValidateAll;
--- a/validator_derive/src/lib.rs
+++ b/validator_derive/src/lib.rs
@@ -26,12 +26,19 @@ mod validation;
 #[proc_macro_error]
 pub fn derive_validation(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     let ast = syn::parse(input).unwrap();
-    impl_validate(&ast).into()
+    impl_validate(&ast, false).into()
 }
 
-fn impl_validate(ast: &syn::DeriveInput) -> proc_macro2::TokenStream {
+#[proc_macro_derive(ValidateAll, attributes(validate))]
+#[proc_macro_error]
+pub fn derive_validation_all(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
+    let ast = syn::parse(input).unwrap();
+    impl_validate(&ast, true).into()
+}
+
+fn impl_validate(ast: &syn::DeriveInput, nest_all_fields: bool) -> proc_macro2::TokenStream {
     // Collecting the validators
-    let mut fields_validations = collect_field_validations(ast);
+    let mut fields_validations = collect_field_validations(ast, nest_all_fields);
     let mut struct_validations = find_struct_validations(&ast.attrs);
     let (arg_type, has_arg) =
         construct_validator_argument_type(&mut fields_validations, &mut struct_validations);
@@ -120,13 +127,16 @@ fn collect_fields(ast: &syn::DeriveInput) -> Vec<syn::Field> {
     }
 }
 
-fn collect_field_validations(ast: &syn::DeriveInput) -> Vec<FieldInformation> {
+fn collect_field_validations(ast: &syn::DeriveInput, nest_all_fields: bool) -> Vec<FieldInformation> {
     let mut fields = collect_fields(ast);
 
     let field_types = find_fields_type(&fields);
     fields.drain(..).fold(vec![], |mut acc, field| {
         let key = field.ident.clone().unwrap().to_string();
-        let (name, validations) = find_validators_for_field(&field, &field_types);
+        let (name, mut validations) = find_validators_for_field(&field, &field_types);
+        if nest_all_fields && validations.len() == 0 {
+            validations.push(FieldValidation::new(Validator::Nested));
+        }
         acc.push(FieldInformation::new(
             field,
             field_types.get(&key).unwrap().clone(),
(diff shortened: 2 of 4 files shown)


Keats:
That should be an option on the current validate macro rather than a complete new impl.
There is a full rewrite ongoing (https://github.com/Keats/validator/pull/262) so that's probably not going to be merged.

cc @pintariching

validator_derive/src/lib.rs
```diff
@@ -208,112 +211,153 @@ fn quote_field_validations(
 }
 
 /// Find if a struct has some schema validation and returns the info if so
-fn find_struct_validation(attr: &syn::Attribute) -> SchemaValidation {
+fn find_schema_validation(path: &syn::Path, nested: &syn::punctuated::Punctuated<syn::NestedMeta, syn::token::Comma>) -> SchemaValidation {
```
davidlang42, line 214: This looks like a huge change but actually its just renaming the function and pulling the first 2 lines of the `if_chain` into the calling function.

validator_derive/src/lib.rs
```diff
@@ -208,112 +211,153 @@ fn quote_field_validations(
...
     let error = |span: Span, msg: &str| -> ! {
         abort!(span, "Invalid schema level validation: {}", msg);
     };
 
-    if_chain! {
-        if let Ok(syn::Meta::List(syn::MetaList { ref nested, .. })) = attr.parse_meta();
-        if let syn::NestedMeta::Meta(syn::Meta::List(syn::MetaList { ref path, ref nested, .. })) = nested[0];
-
-        then {
```
davidlang42, line 220: Everything from here down (in this function) is just whitespace

validator_derive/src/lib.rs
```diff
@@ -208,112 +211,153 @@ fn quote_field_validations(
...
 }
 
 /// Finds all struct schema validations
-fn find_struct_validations(struct_attrs: &[syn::Attribute]) -> Vec<SchemaValidation> {
-    struct_attrs
-        .iter()
-        .filter(|attribute| attribute.path == parse_quote!(validate))
-        .map(find_struct_validation)
-        .collect()
+fn collect_struct_validations(struct_attrs: &[syn::Attribute]) -> (Vec<SchemaValidation>, bool) {
```
davidlang42, line 305: This is the actual change, most code is just for handling the many cases of meta types (similarly to `collect_field_validations`)

davidlang42:
I've rewritten it has a struct attribute (`#[validate(nest_all_fields)]`) rather than new macro, as you suggested.

I understand the rewrite is ongoing, but as it doesn't seem very close to being merged I'm hoping you might consider merging this version (not being pushy though, just putting it out there).

It was actually a fairly small change to make this work (once I understood how the existing derive macro worked) so thanks for structuring well in the first place!

Also, the unit tests were quite helpful in making sure I didn't break this, they all pass now and I've added 2 more to account for the new functionality I've added.

pintariching:
With this change, all the fields have to be nested structs, otherwise the macro calls `.validate()` on any other type. You would need a `#[validate(skip)]` on fields you don't want validated and I don't think it's implemented in the main branch?

I can get this working in the rewrite.

Keats:
> With this change, all the fields have to be nested structs, otherwise the macro calls .validate() on any other type. You would need a #[validate(skip)] on fields you don't want validated and I don't think it's implemented in the main branch?

Yep you would definitely need a skip attribute as well

davidlang42:
> With this change, all the fields have to be nested structs, otherwise the macro calls `.validate()` on any other type. You would need a `#[validate(skip)]` on fields you don't want validated and I don't think it's implemented in the main branch?
> 
> I can get this working in the rewrite.

I included this! I just didn't call it skip. I called it `alwaysvalid` but I can rename to `skip` if you'd rather :)

validator/src/lib.rs
```diff
@@ -90,4 +90,4 @@ pub use traits::{Contains, HasLen, Validate, ValidateArgs};
 pub use types::{ValidationError, ValidationErrors, ValidationErrorsKind};
 
 #[cfg(feature = "derive")]
-pub use validator_derive::Validate;
+pub use validator_derive::Validate;
```
jayvdb, line 93: undesirable whitespace change

jayvdb:
IMO skip is the usual term used for this.

PR #271, opened 2023-09-05
Use `chars().count()` instead of `len()` for `Cow<'a, str>`

future-highway:
Updated `HasLen` implementation for `Cow<'a, str>` so that it matches the other `String` types.


pintariching:
The crate is currently undergoing a major rewrite, check out #262.

Keats:
Can you do the PR on the `next` branch?

PR #272, opened 2023-09-24
Span-based email validation

Finchiedev:
Hello folks, I've been working on a command-line library that uses `validator` to report errors in arguments, and have recently focused on improving the quality of reported error messages. For the invalid email `test@"example.com`, instead of:
```text
Invalid email: test@"example.com
```

I would like to report something like:
```text
Invalid email: invalid character in domain
test@"example.com
     ^
```

To facilitate this, I have created a new function, `validate_email_span`, that should behave exactly the same as `validate_email` except instead of returning a `bool`, it returns a `Result` with additional context; an error enum and span for the first error encountered in the email.

I understand such code may be out of scope for this library, but thought it would be worth sending a patch upstream before considering a different crate - this one is quite nice! This new function, and its supporting types, have been placed inside a `span` module. If such contributions are in scope I would like to explore adding more contextual validators, such as IP addresses.

If there's anything I can do to ease the review process please let me know :)

### DIFF
--- a/validator/src/lib.rs
+++ b/validator/src/lib.rs
@@ -65,9 +65,12 @@
 //! ```
 
 mod display_impl;
+pub mod span;
 mod traits;
 mod types;
 mod validation;
+
+pub use span::email::{validate_email_span, EmailError, EmailValidationError};
 
 #[cfg(feature = "card")]
 pub use validation::cards::validate_credit_card;
--- a/validator/src/span/email.rs
+++ b/validator/src/span/email.rs
@@ -1,0 +1,592 @@
+extern crate alloc;
+
+use alloc::borrow::Cow;
+use core::convert::TryFrom;
+use idna::domain_to_ascii;
+use lazy_static::lazy_static;
+use regex::Regex;
+
+use crate::{validation::ip::validate_ip, HasLen};
+
+// Valid characters, excluding ASCII a-z, A-Z and 0-9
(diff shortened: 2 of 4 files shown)


Keats:
Have a look at https://github.com/Keats/validator/pull/262
This is going to be merged next week so that's probably not going to work cc @pintariching 
I don't know how i feel about having those kind of error messages, they seem way too detailed for 99% of the usecases imo?

Finchiedev:
Thanks for letting me know. I am very happy to rebase my changes on top of #262 if they're considered in-scope for this repository.

I do agree this is too much context for most use cases, however however since `validate_email` still exists (as a wrapper around `validate_email_span`) users can decide for themselves the level of granularity they require. This is similar to how `validate_ip` is already implemented: if users want a boolean, they can use `validate_ip`. If they want a more detailed error, they can use the `IpAddr::from_str` it wraps. There is clearly value to both approaches, I'm just not sure if this library is interested in supporting the more detailed use case.

pintariching:
This could be perhaps behind a feature so that you could select what kind of errors you want?

I think that most users use this crate to validate structs in a CRUD API, so having such detailed error messages is unneeded I think?

However for more complex usages I think this could work behind a feature flag, something like `diagnostics` perhaps?

The macro side would need some touch ups in case we accept this as currently it works with bools only.

PR #274, opened 2023-10-04
Use generics and macros

future-highway:
This also implements #271 on next which is included in the code changes related to `HasLen`, though I don't see how `HasLen` is used in this branch.

Most of the code was eliminated by using generic implementations for `Option<T>` and `&T`.

### DIFF
--- a/validator/src/lib.rs
+++ b/validator/src/lib.rs
@@ -82,7 +82,7 @@ pub use validation::regex::ValidateRegex;
 pub use validation::required::ValidateRequired;
 pub use validation::urls::ValidateUrl;
 
-pub use traits::{HasLen, Validate, ValidateArgs};
+pub use traits::{Validate, ValidateArgs};
 pub use types::{ValidationError, ValidationErrors, ValidationErrorsKind};
 
 #[cfg(feature = "derive")]
--- a/validator/src/traits.rs
+++ b/validator/src/traits.rs
@@ -1,92 +1,4 @@
-use std::{
-    borrow::Cow,
-    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
-    rc::Rc,
-    sync::Arc,
-};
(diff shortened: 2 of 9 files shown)


pintariching:
I don't think the trait `HasLen` is used anywhere. I think I forgot to remove it in the rewrite.

Keats:
So for the length validator, I would like to add an option to the validator for the type of length:

- chars (default)
- utf8 bytes
- utf16 bytes (used in HTML)

So you would call it like:

```
#[validate(length(min = 1)]. 
#[validate(length(min = 1, kind="chars")]   // equal to previous line
#[validate(length(min = 1, kind="utf8_bytes")]
#[validate(length(min = 1, kind="utf16_bytes")]
```

What do people think? Not on this PR though, just to get some feedback while talking about it.

Keats:
And I guess we can remove all the HasLen handling if it's unused

future-highway:
I don't have a strong opinion on the type of length since I'm not actually using this library anymore (I was wrapping the lib in my own validation library and `HasLen` was my main use, so I never took full advantage of the features).

My only thought is that it may be an easy way to make a mistake by forgetting to specify. Maybe having distinct choices (e.g., `length_in_u8`, `length_in_chars`, etc.) would prevent such a mistake. `count` could be used for `Vec` and `[T]` to distinguish them.

As I said though, I'm not using this library currently and it's probably a better idea to take the input of others.

I'll update the PR to delete `HasLen`

future-highway:
PR updated. 

May also want to consider replacing the use of `lazy_static` (see: https://github.com/rust-lang-nursery/lazy-static.rs/issues/214) with `once_cell` which also has a lazy interface. I didn't think I should mix it in this PR.

Keats:
Yes, lazy static will be replaced later, thanks!

PR #301, opened 2024-03-06
Nested collections should not always fail

einarmo:
Nested validators on collections would always fail with an empty list of errors, this fixes that and updates all the collection tests to check that it is possible to create a struct that validates.


brogdogg:
@Keats do you plan to release a patch release with these changes? I hit this today in my own validation.

Keats:
Yes, just want to handle the other known bugs to release everything in one patch

PR #309, opened 2024-03-11
Update documentation for custom validations

divergentdave:
This cleans up some outdated documentation using the old syntax for custom validations.


Keats (approved):
Thanks!

PR #315, opened 2024-03-28
Fix custom functions for optional number types

kyrias:


validator_derive/src/types.rs
```diff
@@ -9,50 +11,32 @@ use syn::{Expr, Field, Ident, Path};
...
+        quote!(i8),
+        quote!(i16),
+        quote!(i32),
+        quote!(i64),
+        quote!(i128),
+        quote!(f32),
+        quote!(f64),
+    ];
+    let mut tys = Vec::with_capacity(number_types.len() * 3);
+    for ty in number_types {
```
Keats, line 33: for loop are still not available in const fn right? :(

Keats (approved):

PR #318, opened 2024-04-05
 Expose validator constraints to end-users using a new trait

kyrias:
Re-implements https://github.com/Keats/validator/pull/238 after the derive macro rewrite.


kyrias:
Ping.

Keats:
Thanks for the ping I had forgotten about it. I'll look at it next week.

Keats (commented):
It's a bit invasive and requires keeping the structs updated to match the parameters of the validators. Do we need the constraints to be fully typed? I'm guessing it's going to be passed to the frontend or some other service anyway in most cases

Can we somehow derive it from the darling attributes directly without having to add to the constraints on each function?

PR #323, opened 2024-04-18
Update README

nopeNoshishi:
Hello! I am a user of this validator crate.

I recently had the opportunity to apply a custom validation. While reading the documentation, I noticed that the way to take arguments and add them to the validation differs from what is described in the README. It seems that the latest 0.18.0 has a different definition due to the implementation changes made by the following PR

- https://github.com/Keats/validator/pull/262

Since this is the case, I decided to rewrite the README and create a PR.
While I made an effort to find related issues and PRs, I needed help locating them. If any are pending or planned, I would appreciate it if they could be addressed. I'm open to closing this if necessary.

I have confirmed that the code in the documentation works with `0.18.0`, but I would appreciate it if you could point out any omissions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Installation:
 
 ```toml
 [dependencies]
-validator = { version = "0.16", features = ["derive"] }
+validator = { version = "0.18", features = ["derive"] }
 ```
 
 A short example:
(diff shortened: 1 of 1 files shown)


Keats (approved):
Looks great, thanks!

PR #329, opened 2024-05-15
Publish AsRegex

flx5:
Hi,
I wanted to add the ValidateRegex trait to a custom struct. However I was unable to do so because the regex parameter of the validate_regex function is declared as impl AsRegex. That trait is not reexported.


Keats:
Thanks!

PR #335, opened 2024-06-30
add devcontainer

### LINKED ISSUE #334: Provide devcontainer config
I want to contribute something but I don't want my mac to be polluted with various runtimes/dependencies on the OS level as I only have little SSD space left.

After contributing, I want to throw away all dependencies by just removing the dev container

spotlesscoder:
fixes #334

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,13 @@
+{
+	"name": "Rust",
+	"image": "mcr.microsoft.com/devcontainers/rust:1-bullseye",
+    "customizations": {
+		"vscode": {
+			"settings": {},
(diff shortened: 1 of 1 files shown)


Keats:
I would rather avoid adding editor specific files in the repo.

spotlesscoder:
devcontainer spec is not editor specific. More and more editors implement this.

There is also a cli to use the devcontainer spec without any IDE for example to use it with vim etc.

The customizations are IDE-specific though but they don't hurt ;)

Keats:
Honestly I'd rather these things not be in the repo otherwise we will end up having nix/guix etc setups. It should just be .gitignored.

PR #339, opened 2024-07-14
feat: add `#[validator(crate = "...")]` for crate alias

matteopolak:
Currently it's not possible to package up the `validator` crate's derives without requiring users to include the dependency locally (cluttering up their deps list if they don't actually use it).

This PR adds an optional `crate` attribute that overrides all references to the crate with a new name.

```toml
validator_renamed = { package = "validator", version = "0.18", features = ["derive"] }
```

```rust
#[derive(validator_renamed::Validator)]
#[validate(crate = "validator_renamed")]
pub struct MyData {
  // ...
}
```

It also works with paths, e.g.

```rust
#[validate(crate = "some::other::location")]
```

### DIFF
--- a/validator_derive/src/lib.rs
+++ b/validator_derive/src/lib.rs
@@ -259,7 +259,7 @@ struct ValidationData {
 }
 
 #[derive(Debug, Clone)]
-struct CrateName {
+pub(crate) struct CrateName {
     inner: Path,
 }
 
--- a/validator_derive/src/lib.rs
+++ b/validator_derive/src/lib.rs
@@ -39,7 +39,7 @@ impl ToTokens for ValidateField {
 
         // Length validation
         let length = if let Some(length) = self.length.clone() {
-            wrapper_closure(length_tokens(length, &actual_field, &field_name_str))
+            wrapper_closure(length_tokens(&self.crate_name, length, &actual_field, &field_name_str))
         } else {
             quote!()
         };
@@ -47,6 +47,7 @@ impl ToTokens for ValidateField {
         // Email validation
         let email = if let Some(email) = self.email.clone() {
             wrapper_closure(email_tokens(
+                &self.crate_name,
                 match email {
                     Override::Inherit => Email::default(),
                     Override::Explicit(e) => e,
@@ -61,6 +62,7 @@ impl ToTokens for ValidateField {
         // Credit card validation
         let card = if let Some(credit_card) = self.credit_card.clone() {
             wrapper_closure(credit_card_tokens(
+                &self.crate_name,
                 match credit_card {
                     Override::Inherit => Card::default(),
                     Override::Explicit(c) => c,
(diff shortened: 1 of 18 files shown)


Keats:
Is it possible to add a test?

matteopolak:
Added (also fixed the compile error). The tests cover:
- Accidentally using `::validator::...` instead of `#crate_name::...` (by the compile-fail)
- Accidentally using `validator::...` instead of `#crate_name::...` (by the run-pass)

It doesn't cover every branch, but I checked the ones for `url`, `email`, `length`, `range`, and `custom` (function)

matteopolak:
Whoops, seems like that "private" type is now permitted on the nightly channel I was on, but not on 1.70. I shuffled it around so it could be `pub` to work around the issue, and tested it on `1.70` (test + build).

Keats:
It looks like ti still fails?

matteopolak:
The fail appears to be unrelated to this commit and more related to the version of `nightly` being used having changed the order of the `note` diagnostics:

```
EXPECTED:
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
error: Invalid argument reference

         = note: The lifetime `'a` is not supported.
         = help: Please use the validator lifetime `'v_a`

 --> tests/compile-fail/custom/different_lifetime.rs:8:22
  |
8 | #[validate(context = "Arg<'a>")]
  |                      ^^^^^^^^^
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈

ACTUAL OUTPUT:
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
error: Invalid argument reference
 --> tests/compile-fail/custom/different_lifetime.rs:8:22
  |
8 | #[validate(context = "Arg<'a>")]
  |                      ^^^^^^^^^
  |
  = note: The lifetime `'a` is not supported.
  = help: Please use the validator lifetime `'v_a`
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
```

Looks like it has placed them at the end instead of at the start for a lot of the expect-fail tests. It might be a good idea to pin to the last-working nightly, or maybe there's a way to specify different output for a different toolchain?

Keats:
I think I should remove nightly entirely from the CI, it's always failing on there for one reason or another

matteopolak:
Sounds like a good idea. Let me know if you need anything from me to get this merged!

PR #342, opened 2024-07-21
Fix incorrect example in README

naskya:
This pull request fixes a minor error in README where `#[validate(nested)]` is written as `#[validate]`.


Keats:
Thanks!

PR #348, opened 2024-08-14
Add ValidateRegex Implementation For Box, Rc, and Arc

Kailokk:
Added a blanket implementation where the inner type T for these container types implements `ValidateRegex`.

These types are sometimes used in place of `String` where the mutability that `String` provides is not necessary.

### DIFF
--- a/validator/src/validation/regex.rs
+++ b/validator/src/validation/regex.rs
@@ -102,6 +102,12 @@ impl ValidateRegex for &str {
     }
 }
 
+impl ValidateRegex for str {
+    fn validate_regex(&self, regex: impl AsRegex) -> bool {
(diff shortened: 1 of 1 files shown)


Kailokk:
> Looks good. Do you run into the same issue with other validators?

I haven't so far, i've been hoping to to take a proper look when i had the chance, this one just came up in something i was working on.

Keats (approved):

Kailokk:
@Keats thanks for the approval. I don't have the permissions to enact the merge, so i will leave that to you?

Keats:
Yep that's good, I'll merge/release soon

Kailokk:
Awesome, thank you so much for your time, and this ever useful library

PR #350, opened 2024-09-06
Swap to Proc Macro Error 2 RUSTSEC-2024-0370

pvichivanives:
Hi! 
Given the RUSTSEC warning issued here: https://rustsec.org/advisories/RUSTSEC-2024-0370 just submitting a PR to move to the reccomended new version of procmacroerror2
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,5 +24,5 @@ jobs:
         run: |
           cargo build --no-default-features
           cargo test --no-default-features
-          cargo build --all-features
-          cargo test --all-features
+          cargo build --features unic --features derive --features card
+          cargo test  --features unic --features derive --features card
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,5 +47,5 @@ jobs:
         run:
           cargo build --no-default-features
           cargo test --no-default-features
-          cargo build --features all
-          cargo test  --features all
+          cargo build --all-features
+          cargo test  --all-features
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Build System Info
         run: rustc --version
       - name: Tests
-        run:
+        run: |
           cargo build --no-default-features
           cargo test --no-default-features
           cargo build --all-features
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Install Rust
-        uses: dtolnay/rust-toolchain@stable
+        uses: dtolnay/rust-toolchain@nightly
         with:
           toolchain: ${{ matrix.rust }}
       - name: Cache dependencies
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,3 +26,26 @@ jobs:
           cargo test --no-default-features
           cargo build --features unic --features derive --features card
           cargo test  --features unic --features derive --features card
+  test_validator-nightly:
+    name: Continuous integration
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - rust: nightly
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: ${{ matrix.rust }}
+      - name: Cache dependencies
+        uses: Swatinem/rust-cache@v2
+      - name: Build System Info
+        run: rustc --version
+      - name: Tests
+        run:
+          cargo build --no-default-features
+          cargo test --no-default-features
+          cargo build --features all
+          cargo test  --features all
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Install Rust
-        uses: dtolnay/rust-toolchain@nightly
+        uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: ${{ matrix.rust }}
       - name: Cache dependencies
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         include:
-          - rust: 1.79.0 # MSRV, sometimes it changes because of compiler test issues
+          - rust: 1.81.0 # MSRV, sometimes it changes because of compiler test issues
           - rust: stable
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,4 +48,4 @@ jobs:
           cargo build --no-default-features
           cargo test --no-default-features
           cargo build --all-features
-          cargo test  --all-features
+          cargo test --all-features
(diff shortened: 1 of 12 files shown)


jayvdb:
Please run the  tests with `TRYBUILD=overwrite` set.

jayvdb:
ping @Keats

pvichivanives:
Apparently the error messages between 1.79 and 1.81 are different for procmacroerror 2

validator_derive/Cargo.toml
```diff
@@ -17,6 +17,6 @@ proc-macro = true
 syn = "2"
 quote = "1"
 proc-macro2 = "1"
-proc-macro-error = "1"
+proc-macro-error2 = "2"
```
HigherOrderLogic, line 20: Maybe consider adding a "nightly" feature to the `validator_derive` crate?

Since [automatic nightly detection](https://github.com/GnomedDev/proc-macro-error-2/commit/e4e879d7def0b0a8910c7de0d7bdbedf8018795f) has been removed, and the "nightly" feature is disabled by default, some running with nightly compiler wont have nightly warnings.

pvichivanives: Thanks! Added

validator_derive/Cargo.toml
```diff
@@ -17,6 +17,9 @@ proc-macro = true
...
 darling = { version = "0.20", features = ["suggestions"] }
 once_cell = "1.18.0"
+
+[features]
+nightly = ["proc-macro-error2/nightly"]
```
Keats, line 25: can the name be more clear somehow?

validator_derive/Cargo.toml
HigherOrderLogic: You forget to add the nightly feature to the main crate too. Users wont use this crate directly but rather through the main `validator` crate.

validator_derive/Cargo.toml
pvichivanives: nightly_features?

validator_derive/Cargo.toml
pvichivanives: Yup oops

validator_derive/Cargo.toml
Keats: I'm not sure, I don't even know what this does in proc-macro-error2

pvichivanives: For the nightly feature in proc-macro-error2 its just a different implementation for emitting diagnostics and aborting. The main difference between the two appear to be that warnings are emitted in nightly while ignored in non-nightly builds. Tagging @HigherOrderLogic since they requested this


If you want to look at the code its here: https://github.com/GnomedDev/proc-macro-error-2/tree/master/src/imp 
The delegate.rs is the nightly version and fallback.rs is the non nightly version

HigherOrderLogic: > I'm not sure, I don't even know what this does in proc-macro-error2

It's as @pvichivanives said, this feature enables emitting warning when using nightly compiler, otherwise ignored.

So to match the function, I think this feature can be renamed to `nightly-warnning`, or `nightly-proc-macro`, or even the longer `nightly-proc-macro-error`

domenicquirl: I've written this in https://github.com/Keats/validator/issues/351, but I'd like to point it out here too: `proc-macro-error2` fundamentally does not do the same thing as `proc-macro-error`, which is to _avoid_ the need to have the user switch on a separate crate feature to get better proc macro diagnostics on nightly. The entire point of `proc-macro-error` is that both sets of error reporting functionality (via `compiler_error!` or via the nightly proc macro diagnostics API) are available to proc macro authors already, but instead of the library author having to decide (or add a feature), whether the nightly APIs can be used is detected at build time via `build.rs`. Using `proc-macro-error2` still provides a shim around `#[cfg(nightly_macro_diagnostics)]` so there's a single API for the macro code no matter the compiler version, but doesn't do the auto-detection.

If this is fine for you @Keats, then this might not be a concern for this PR, but you should make sure that this is actually what you want. In particular, I've linked a different crate in the issue above (`proc-macro2-diagnostics`) that's a similar kind of drop-in replacement but which _does_ do the build-time compiler detection. To me, this would definitely be preferable because IMO the point of using `proc-macro-error` is to _automatically_ give good error messages where possible. At minimum, if you decide to go forwards with this PR and `proc-macro-error2`, the feature should receive some documentation in the crate docs. Otherwise, no one's gonna know what it does or turn it on, because most even nitghtly users will probably not know what a `proc-macro-error2` is or what it does and that they could be getting much better error messages (which again is why I think it would be better if they didn't have to know at all).

Keats: Honestly I don't really care which version gets used. The least disturbance the better though. If someone took proc-macro-error and just bumped syn without any other changes it would be fine. But I agree that needing a feature for some nightly stuff is not great

pvichivanives: Happy to swap over the pr/close this pr if you would want to make one @domenicquirl. Am slightly concerned though that `proc-macro2-diagnostics` has 1 commit in the last 12 months, being a version bump which hasn't been released yet (last release was around 1year ago). Also https://github.com/SergioBenitez/version_check/pull/23 <- for this as a rust team dev pointed out they would prefer people didn't automatically detect nightly.

domenicquirl: I wouldn't be too worried about that, Sergio has been active in the issues over there as recent as a month ago. It's just a small crate that doesn't need lots of updating. 

Aside from that, point taken on nightly detection in general. It's maybe worth pointing out that there's also [a PR](https://github.com/SergioBenitez/proc-macro2-diagnostics/pull/10) to `proc-macro2-diagnostics` that tries to put their nightly detection (the detection itself, not the usage of the nightly APIs) behind a feature flag.

Personally, I don't use `validator` in a nightly context, which means I don't get nice diagnostics either way, so I probably won't push for a specific alternative solution as long as the `syn` version gets bumped with the switch. Just wanted to make sure that the change in behaviour is known, and as I said above I think you should add a section to the top-level crate docs that explains that this feature exists and what it does. Otherwise I'm pretty sure most people will miss it or see it and have a hard time figuring out what it does.

pvichivanives:
Interesting, its defenitely all passing locally except for the error message issue that I mentioned earlier. IS CI turning all features to true?

Keats:
It does: https://github.com/Keats/validator/blob/master/.github/workflows/ci.yml#L25-L28

pvichivanives:
TY! Haven't used the github workflow before so didn't know ci was there. Updated it to only run all the feature flags. Still that issue though of the `TRYBUILD=overwrite` set where rn its set to 1.70. I can set it to 1.81 as well but that then we would fail. The wording was changed in 1.81. So I also bumped the tests MSRV to 1.81 but I can rever that commit if not desired.

.github/workflows/ci.yml
```diff
@@ -24,5 +24,5 @@ jobs:
         run: |
           cargo build --no-default-features
           cargo test --no-default-features
-          cargo build --all-features
-          cargo test --all-features
\ No newline at end of file
+          cargo build --features unic --features derive --features card
+          cargo test  --features unic --features derive --features card
```
Keats, line 28: What's the difference between those diffs?

Keats: Can you add a nightly CI run that enables all the features? Otherwise that feature is never tested

Keats:
That looks ok to me. Ok for everyone else as well?

pvichivanives:
Wanted to note that we have an issue with Workflow though in that Nightly uses different highlighting to stable. 
image

Keats:
Maybe just try to build in CI using nightly rather than run the test?

pvichivanives:
Happy to just build in CI too, I thought it might be more useful if we at least run all the tests we can though.

Keats (approved):

PR #357, opened 2024-11-08
Issue: Panic when trying to use custom validators combined with nested validators

gonzedge:
While trying to use a combination of nested and custom validators, I've been getting `panic`s with the message `Attempt to replace non-empty ValidationErrors entry`. 

I've boiled down the failing case to this minimum setup (contained within this PR): https://github.com/Keats/validator/blob/1728fd7a6b17a200141b820f5bc7e6fc5088ff65/validator_derive_tests/tests/nested.rs#L85-L130

I'm trying to get familiar with the code, but have been unable to correct the issue so far. Any help would be appreciated.

---

Full panic stack backtrace:

```
Attempt to replace non-empty ValidationErrors entry
thread 'fails_nested_validation_multiple_members_and_custom_validations' panicked at validator/src/types.rs:185:13:
Attempt to replace non-empty ValidationErrors entry
stack backtrace:
   0: rust_begin_unwind
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/std/src/panicking.rs:662:5
   1: core::panicking::panic_fmt
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:74:14
   2: validator::types::ValidationErrors::add_nested
             at <redacted>/validator/validator/src/types.rs:185:13
   3: validator::types::ValidationErrors::merge_self
             at <redacted>/tools/validator/validator/src/types.rs:80:21
   4: <nested::fails_nested_validation_multiple_members_and_custom_validations::Root as validator::traits::ValidateArgs>::validate_with_args
             at ./tests/nested.rs:87:14
   5: <nested::fails_nested_validation_multiple_members_and_custom_validations::Root as validator::traits::Validate>::validate
             at ./tests/nested.rs:87:14
   6: nested::fails_nested_validation_multiple_members_and_custom_validations
             at ./tests/nested.rs:125:18
   7: nested::fails_nested_validation_multiple_members_and_custom_validations::{{closure}}
             at ./tests/nested.rs:86:69
   8: core::ops::function::FnOnce::call_once
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/ops/function.rs:250:5
   9: core::ops::function::FnOnce::call_once
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/ops/function.rs:250:5
```

PR #359, opened 2024-11-11
Fix Merge issues between custom and nested

pvichivanives:
A response that should fix #357 
Did not include the test in this PR since it should be merged in from #357 

Slight behavior change in that it we dont run nested if custom has an error. its a bug fix should be called a patch/minor? (Given that this behavior is better than a panic)

Called an unwrap here to show error output from #357 
image

### DIFF
--- a/validator/src/types.rs
+++ b/validator/src/types.rs
@@ -181,10 +181,9 @@ impl ValidationErrors {
     fn add_nested(&mut self, field: &'static str, errors: ValidationErrorsKind) {
         if let Vacant(entry) = self.0.entry(field) {
             entry.insert(errors);
+        } else {
+            panic!("Attempt to replace non-empty ValidationErrors entry");
         }
-        // Else case here is simply do nothing. Pretty sure this is caught earlier to not run but since
-        // If we get here, just not adding it does the same thing (although waste extra compute)
-        // probably better to just ignore here.
     }
 
     #[must_use]
--- a/validator/src/types.rs
+++ b/validator/src/types.rs
@@ -181,9 +181,10 @@ impl ValidationErrors {
     fn add_nested(&mut self, field: &'static str, errors: ValidationErrorsKind) {
         if let Vacant(entry) = self.0.entry(field) {
             entry.insert(errors);
-        } else {
-            panic!("Attempt to replace non-empty ValidationErrors entry");
         }
+        // Else case here is simply do nothing. Pretty sure this is caught earlier to not run but since
+        // If we get here, just not adding it does the same thing (although waste extra compute)
+        // probably better to just ignore here.
     }
 
     #[must_use]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,8 @@
 ## Changelog
 
-
 ## Unreleased
 - Implement `AsRegex` for `std::sync::LazyLock<Regex>`
 - Bug fix for nested issue with custom only running nested if outer passes
-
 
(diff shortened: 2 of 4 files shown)


pvichivanives:
One comment here is it could make sense to join it in a list rather than the map we currently have with nested/self. Would require more logic but It shouldn't be too difficult. Would like thoughts on this though so I put up the simplest clean solution

Keats:
I think it's fine

validator/src/types.rs
```diff
@@ -179,10 +179,16 @@ impl ValidationErrors {
...
-            panic!("Attempt to replace non-empty ValidationErrors entry");
+        match self.0.get(field) {
+            None => {
+                self.0.insert(field, errors);
+            }
+            Some(entry) => {
+                let mut new_table = ValidationErrors::new();
+                new_table.0.insert("self", entry.clone());
+                new_table.0.insert("nested", errors);
+                self.0.insert(field, ValidationErrorsKind::Struct(Box::new(new_table)));
```
Keats, line 190: I think rather than changing that, it would be better to redesign the whole error types so you only get 1 breaking change instead of 2 with that new change

pvichivanives: Would need a bit more explanation on how a redesign would cause less breaking changes. I don't see how this would cause 2 breaking changes? (I would argue that we can consider this 0 breaking changes for SemVer since all this does is fixes a panic bug case which mean the current behavior when this exact case happens is undefined?)

Keats: Right now it panics so it doesn't work and people don't rely on it. If we merge that, people might start to rely on the self/nested fields which would be one more breaking change compared to if we did not merge that. 
I'll try to sketch what the new errors would look like this afternoon

pvichivanives: Oh you mean in the future if we change. Got it. I thought you meant 2 breaking changes with this PR

gonzedge: @Keats what do you suggest as our path forward here?

Keats: Looking at https://github.com/Keats/validator/pull/357, we should basically not be in a position to have both custom and nested.

In 
```
         #[validate(nested, custom(function = "all_value1s_are_unique"))] 
         a: Vec<A>, 
```
we should run the custom function first and only if it passes run the nested validations

pvichivanives: Sorry for delay, I think this should be done now. I have the extra case where if it is missed and we did run the nested and custom has errors, we ignore the nested too.

validator/src/types.rs
```diff
@@ -181,9 +181,10 @@ impl ValidationErrors {
     fn add_nested(&mut self, field: &'static str, errors: ValidationErrorsKind) {
         if let Vacant(entry) = self.0.entry(field) {
             entry.insert(errors);
-        } else {
-            panic!("Attempt to replace non-empty ValidationErrors entry");
         }
+        // Else case here is simply do nothing. Pretty sure this is caught earlier to not run but since
+        // If we get here, just not adding it does the same thing (although waste extra compute)
+        // probably better to just ignore here.
```
Keats, line 187: I think the panic is helpful to have in case we are missing cases. At least it's going to point out another bug

Keats (commented):
Maybe just keep the panic?

validator/src/types.rs
pvichivanives: Sure! I'll add the panic back.

Issue #141, opened 2021-01-22
Add not_match validator


mattremmel:
This would be nice to have for me, specifically for cases where all inputs are valid except for a single case (Uuid::nil()), or where two fields can contain similar types of data, but they can't exactly match. I haven't spent much time looking at the internals, but i'd be willing to try to write the code if its feasible to do.

xFrednet:
I like the idea if negating a validation. This would be the negation of the `match` validation if I'm right. The question is what syntax could be used. Do you have a suggestion how to write this? @mattremmel :upside_down_face:

mattremmel:
I was originally thinking just an inverted version of `must_match`, but as I think about it, I'm not sure if that is needlessly restrictive. Its probably possible to do a generic `not(...)`, that would would allow for things like `#[validate(not(length(min =1, max = 3)))]`, the downside being another level of parenthesis around everything, but gaining increased flexibility.

Keats:
I'm not sure we want to add generic `not`. `match` is one that might need it but the other validator seem to not need it.  I mean, how often do you need something like `#[validate(not(length(min =1, max = 3)))]`

IniterWorker:
I second @Keats. This is not the way to go. This could be solved with a custom `validator`.

PR #365, opened 2025-01-21
Add must_not_match validator

Jawell:
Hi there! I want to add one more validator if you don't mind


Jawell:
@Keats ping :)

validator_derive/src/lib.rs
```diff
@@ -178,6 +179,19 @@ impl ToTokens for ValidateField {
             quote!()
         };
 
+        // Must not match validation
+        let must_not_match = if let Some(must_not_match) = self.must_not_match.clone() {
+            // TODO: handle option for other
```
Keats, line 184: Is that still a TODO?

Jawell: As I understood correctly - it was handled

validator/src/validation/must_not_match.rs
```diff
@@ -0,0 +1,63 @@
+/// Validates that the 2 given fields do not match.
+/// Both fields are optionals
+#[must_use]
+pub fn validate_must_not_match<T: Eq>(a: T, b: T) -> bool {
```
Keats, line 4: do we want Eq or PartialEq?

Jawell: Hmm, should be PartialEq

Keats:
Looks like legit failures

Jawell:
> Looks like legit failures

I fixed tests

Jawell:
@Keats reminder :)

README.md
```diff
@@ -212,8 +212,10 @@ const MIN_CONSTANT: i32 = 0;
 
 ```
 
-### must_match
-Tests whether the 2 fields are equal. `must_match` takes 1 string argument. It will error if the field
+### must_match and must_not_match
```
Keats, line 215: I'm wondering if we can use `must_match` rather than create a new validator. Something like `must_match(expected=false)` with a default of true (bad name but just to illustrate)

Jawell: As you didn't approve generic `not()` ([Add not_match validator #141](https://github.com/Keats/validator/issues/141#issuecomment-766959104)) and we have `contains` and `does_not_contain` I thought that we need a new one

Jawell: So, can we move on?

Jawell:
@Keats no problem
But only one question: how to access other fields in custom validators? I didn't find anything

Keats:
You can't right now, you have to do a schema level validation fn

PR #366, opened 2025-01-29
Remove dependency on unic-ucd-common by using std::char::is_control

mmalecot:
Checking whether a character is a control character is already present in the `std`, so no external dependency is required and the `unic` feature no longer needs to exist (or at least is built-in).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,8 +24,8 @@ jobs:
         run: |
           cargo build --no-default-features
           cargo test --no-default-features
-          cargo build --features unic --features derive --features card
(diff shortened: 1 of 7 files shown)


Keats (approved):

mmalecot:
Yes, I use it to ensure that a string contains only single-line text for example.

AverageHelper:
Since `unic-ucd-common` and others are beginning to throw warnings in cargo-audit, I'd very appreciate a release for this to crates.io :pray:

Keats:
oh wow i didn't realise i didn't release a new version of validator at all

PR #378, opened 2025-04-14
Feature flag out the external dependencies

rosscarter154:
The rationale for this change is that there are cases where people would like to use the general validation infrastructure, as well as the validation derive macros for simple data types, but do not want to pull in these extra dependencies to their dependency graph. The remaining validators that are not affected by this PR affect built-in types and as such are always enabled.

By enabling the new feature flags by default, this is **not** a breaking change, unless the downstream user explicitly set default-features = false.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,12 +155,12 @@ validation functions and types.
 ## Validators
 The crate comes with some built-in validators and you can have several validators for a given field.
 
-### email
+### email (optional)
 Tests whether the String is a valid email according to the HTML5 regex, which means it will mark
 some esoteric emails as invalid that won't be valid in a `email` input as well.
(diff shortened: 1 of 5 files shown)


Keats:
> By enabling the new feature flags by default, this is not a breaking change, unless the downstream user explicitly set default-features = false.

I would consider that a breaking change, which is fine for a new version

rosscarter154:
> > By enabling the new feature flags by default, this is not a breaking change, unless the downstream user explicitly set default-features = false.
> 
> I would consider that a breaking change, which is fine for a new version

Bumped the version accordingly and added documentation for the feature flags.

Keats (approved):

PR #379, opened 2025-04-18
Ignore fields with validation instruction with `nest_all_fields`

LeMarwin:
Problem:
* `nest_all_fields` adds #[validate(nested)] for all fields, making the macro unusable in a situation where some fields have specific validation instruction.
* adding `#[validate(skip)]` along with other `#[validate]` attrs skips field validation entirely

Solution: only add `#[validate(nested)]` to fields that don't have other validation attributes.

Essentially this makes `nest_all_fields` semantically closer to `validate_all_fields` but inserts `#[validate(nested)]` if no validation instruction is present. But renaming it is a breaking change

Concrete example of what I wanted to enable:

```rust
#[derive(Validate)]
#[validate(nest_all_fields)]
struct Test {
    #[validate(skip)]
    _a: Nested,
    b: NestedValidated,
    #[validate(range(min = 10))]
    c: u8,
}
```

It's obvious that this should work: we gave explicit validation instruction, so there's no need to insert `#[validate(nested)]`

Moreover, this PR enables something like this:

``` rust
#[derive(Validate)]
#[validate(nest_all_fields)]
struct Test {
    #[validate(skip)]
    _a: Nested,
    b: NestedValidated,
    #[validate(custom(function = "validate_c"))]
    c: NestedValidated,
}
```

which follows the same logic as the previous example, and is quite handy in general

### DIFF
--- a/validator_derive/src/lib.rs
+++ b/validator_derive/src/lib.rs
@@ -336,7 +336,9 @@ pub fn derive_validation(input: proc_macro::TokenStream) -> proc_macro::TokenStr
             validation_fields = validation_fields
                 .iter_mut()
                 .map(|f| {
+                    if !f.has_validate_instruction() {
                     f.nested = Some(true);
+                    }
                     f.to_owned()
                 })
                 .collect();
--- a/validator_derive/src/types.rs
+++ b/validator_derive/src/types.rs
@@ -219,6 +219,27 @@ impl ValidateField {
             ),
         }
     }
+
+    /// Check to see if the field has validation instruction
+    pub fn has_validate_instruction(&self) -> bool {
+        [
+            self.credit_card.is_some(),
+            self.contains.is_some(),
+            self.does_not_contain.is_some(),
+            self.email.is_some(),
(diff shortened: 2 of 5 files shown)


Keats:
How about things like that https://github.com/Keats/validator/blob/master/validator_derive_tests/tests/nested.rs#L273-L336 eg containers where you have length validation but also want to validate the values

LeMarwin:
> How about things like that https://github.com/Keats/validator/blob/master/validator_derive_tests/tests/nested.rs#L273-L336 eg containers where you have length validation but also want to validate the values

Okay, I missed that usecase since tests didn't break.

I guess that makes this PR a breaking change and thus undesirable, since if someone relied on `#[validate(nest_all_fields)]` to insert `#[validate(nested)]` for containers that would no longer be the case and this change would be silent.

What about adding `validate_all_fields` as a container-level attribute? So that the compiler actually checks that all fields have some kind of validation instruction and adds `#[validate(nested)]` if none is present?

PR #380, opened 2025-05-03
Fix email validation to reject TLD-only domains like user@com

Lucas-BRT:
This pull request enhances the email validation logic in `validator/src/validation/email.rs` to ensure stricter compliance with domain validation rules. It also updates the corresponding test cases to reflect these changes and adds new tests to cover edge cases.

### Email Validation Logic Updates:

* Modified `validate_domain_part` to reject email addresses with top-level domains (e.g., `user@com`) by requiring at least one dot in the domain part. This ensures stricter validation of domain structures.

### Test Case Updates:

* Updated an existing test case to mark `abc@bar` as invalid, aligning with the stricter domain validation rules.
* Added a new test case `test_user_at_com_fails` to validate that `user@com` is correctly identified as an invalid email address.
* Introduced a test in `validator_derive_tests/tests/email.rs` to ensure that custom implementations for email validation also reject top-level domain-only email addresses. This includes detailed error assertions to verify the validation logic.

### DIFF
--- a/validator/src/validation/email.rs
+++ b/validator/src/validation/email.rs
@@ -19,20 +19,21 @@ static EMAIL_LITERAL_RE: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"\[([a-fA-F0-9:\.]+)\]\z").unwrap());
 
 /// Checks if the domain is a valid domain and if not, check whether it's an IP
-#[must_use]
 fn validate_domain_part(domain_part: &str) -> bool {
-    if EMAIL_DOMAIN_RE.is_match(domain_part) {
+    // if it's a domain literal like [127.0.0.1]
+    if let Some(caps) = EMAIL_LITERAL_RE.captures(domain_part) {
+        return match caps.get(1) {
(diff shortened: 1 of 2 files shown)


Keats:
Technically user@com can be a valid email. The current regex matches the HTML5 email regex that does allow it

Lucas-BRT:
Although HTML5 allows email addresses like user@com with its internal regex, this does not reflect the stricter rules required by actual email validation standards.

HTML5 validation, as shown in W3Schools, does not allow user@com.

image


Allowing this could lead to issues with DNS systems and other technical risks, as outlined in ICANN's guide on dotless domains.

For more robust validation, stricter rules should be applied.

RFC 5321: https://datatracker.ietf.org/doc/html/rfc5321

RFC 5322: https://datatracker.ietf.org/doc/html/rfc5322

W3Schools: https://www.w3schools.com/tags/att_input_pattern.asp

ICANN Guide: https://www.icann.org/en/announcements/details/new-gtld-dotless-domain-names-prohibited-30-8-2013-en

PR #393, opened 2026-04-14
Update to rust 1.86, fix ci, introduce clippy and fmt

Th3Link:
Since the ci was not working when reviewing a PR, I tried to make it working again.

Root cause:
indexmap >=2.12 goes for hashbrown 0.17 which needs rust 1.86
idna -> idna_adapter -> icu... also needs rust 1.86

I see no other way then to go along with it. This also obsolets #385.

I also introduced clippy and rustfmt check since i noticed wrong formatting and linter warings. I also fixed them they where only minor changes.


Keats:
Thanks!

Issue #396, opened 2026-06-08
The `proc-macro-error2` is unmaintained (RUSTSEC-2026-0173)


goofacz:
The `validator_derive` depends on it. Do you plan to replace that dependency?

The RUSTSEC suggests `manyhow` or `proc-macro2-diagnostics`.

gamma0987:
The RUSTSEC advisory lists `manyhow` and `proc-macro2-diagnostics` as alternatives, but both require rewriting how errors are reported in your proc macros. [`proc-macro-error3`](https://github.com/gamma0987/proc-macro-error3) is a maintained fork and **drop-in replacement** — same API, no code changes beyond the dependency rename, so the switch is minimal and low-risk.

I maintain `proc-macro-error3` and use it in my projects, mainly https://github.com/gungraun/gungraun. Would you be open to this switch? I'm happy to submit a PR if that would help.

jilvin:
Maybe someone can take up the `proc-macro-error2` project or use `proc-macro-error3` as mentioned by @gamma0987 after some code review?

gamma0987:
@jilvin Not sure if I understood your comment right but proc-macro-error3 is the successor to proc-macro-error2

ethteck:
I'd be grateful for a new release with the change to proc-macro-error3 a few weeks ago.

Keats:
Sorry forgot about it. I've pushed 0.20.1 to crates.io

denislituev:
> Sorry forgot about it. I've pushed 0.20.1 to crates.io

Hi! The version on crates.io is still 0.20.0. Could you publish a new one please.

Keats:
It's validator_derive that got bumped, not validator

PR #397, opened 2026-06-10
Replace proc-macro-error2 with proc-macro-error3

denislituev:
proc-macro-error2 is unmaintained. 
proc-macro-error3 is a maintained fork with the same API and several bug fixes including a future-incompatibility warning from rustc.


denislituev:
#396

Keats:
Thanks!

